PR #1, opened 2018-05-29
Model context manager, primitive default sampling, random variable class

sharanry:
Test Notebook : https://github.com/sharanry/pymc4/blob/592c5384de053d54a97427c7478c72426e634b41/test.ipynb

- Currently the Model class uses `treedict` to keep track of the variables defined under it. Exploring whether `treedict` is actually required or python dictionaries can replace this.
- Will explore the use of contextlib 
- Exploring the use of existing pymc3 samplers.


ferrine:
The reason treedict is used in pymc3 is model nesting. Once we were thinking about building blocks that can contan variables. Model definition would come in `__init__` like in regular context manager

pymc4/model.py
```diff
@@ -0,0 +1,129 @@
...
+            return (dict.__contains__(self, item) or
+                    self.parent.tree_contains(item))
+        elif isinstance(self.parent, dict):
+            return (dict.__contains__(self, item) or
+                    self.parent.__contains__(item))
+        else:
+            return dict.__contains__(self, item)
+################################
+
+class Model(Context):
```
ferrine: https://github.com/pymc-devs/pymc3/blob/master/pymc3/model.py#L278
This code allowed model nesting, you might add this later, without it, treedicts etc are useless

twiecki:
@ferrine We never used model nesting and I think we should drop it.

pymc4/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .model import Model
+from .random_variable import *
```
ColCarroll: for `import *` statements, the file should define an `__all__` variable.  This prevents the namespace from getting full of unwanted stuff (right now `from pymc4 import ed` would work, because `random_variable.py` imports edward).

pymc4/model.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def withparent(meth):
+    """Helper wrapper that passes calls to parent's instance"""
+    def wrapped(self, *args, **kwargs):
+        res = meth(self, *args, **kwargs)
+        if getattr(self, 'parent', None) is not None:
+            getattr(self.parent, meth.__name__)(*args, **kwargs)
+        return res
+    # Unfortunately functools wrapper fails
```
ColCarroll: i am not sure this is true any more - we should have a test that fails without `withparent`, or else get rid of it

pymc4/model.py
```diff
@@ -0,0 +1,126 @@
...
+            cls.contexts.stack = []
+        return cls.contexts.stack
+
+    @classmethod
+    def get_context(cls):
+        """Return the deepest context on the stack."""
+        try:
+            return cls.get_contexts()[-1]
+        except IndexError:
+            raise TypeError("No context on context stack")
```
ColCarroll: This could be more informative: "This must be called from inside a `Model` context manager"

pymc4/model.py
```diff
@@ -0,0 +1,126 @@
...
+def modelcontext(model):
+    """return the given model or try to find it in the context if there was
+    none supplied.
+    """
+    if model is None:
+        return Model.get_context()
+    return model
+
+
+def plot(trace):
```
ColCarroll: this function does not need to be here, I think!

pymc4/random_variable.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class RandomVariable(ed.RandomVariable):
+
+    def __init__(
+                self,
+                distribution,
+                sample_shape=(),
+                value=None,
+                name="RV"
```
ColCarroll: the name should probably be the first argument, and have no default argument, right?

ColCarroll:
Ooh, if we drop model nesting, a lot of this code could be quite simplified.  I'd still suggest merging first, adding some unit tests, then coming back to refactor and remove the nesting.

fonnesbeck:
I suggest writing tests earlier rather than later. Doesn’t have to be TDD, but things will get complicated quickly, so best to have them from the start.

pymc4/model.py
sharanry: I am planning to remove all dependency treedict. This shouldn't be even required once I do that.

pymc4/model.py
```diff
@@ -0,0 +1,120 @@
+import tensorflow as tf
+import seaborn as sns
```
fonnesbeck: Don’t need seaborn here

sharanry:
There is some problem while installing tensorflow-probability in travis build. Looking into it.

requirements.txt
```diff
@@ -0,0 +1,6 @@
+xarray==0.10.4
+numpy==1.14.3
+tqdm==4.23.3
+tf-nightly==1.8.0.dev20180331
```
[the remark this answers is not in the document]
sharanry: I tried this. But some imports in tensorflow-probability fails with all latest releases. This is the only combination of versions (for now) in which `import tensorflow-probability` succeeds without any errors.

test/test_model.py
```diff
@@ -0,0 +1,52 @@
+import tensorflow_probability as tfp
+import pymc4 as pm
+from pymc4.random_variable import RandomVariable as RV
+tfd = tfp.distributions
```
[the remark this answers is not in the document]
sharanry:
```
>>> import tensorflow_probability as tfp
>>> import tfp.distributions as tfd
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'tfp'
```

test/test_nothing.py
```diff
@@ -1,5 +1,2 @@
-from pymc4 import *  # pylint: disable=wildcard-import
-
-
 def test_nothing():
     assert 1 == 1
```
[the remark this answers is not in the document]
sharanry: Okay

.pylintrc
```diff
@@ -54,7 +54,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=missing-docstring
+disable=missing-docstring, C0103
```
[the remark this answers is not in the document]
sharanry:
```
tfd = tfp.distributions
tfb = tfp.distributions.bijectors
ed = tfp.edward2
```
Since I am using this and not using all-caps, its causing errors. I am unable to find a different way to import tfd, tfb and ed. Should I selectively ignore it for this page?

sharanry:
@ColCarroll I will be adding more tests for `random_variable` today.

requirements.txt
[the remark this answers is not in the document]
junpenglao: This is issues on the tfp side that are waiting to be resolved - I think for now we can pin the version but just remember to keep updating them by hand.

sharanry:
@ColCarroll Can I merge this?

pymc4/model.py
```diff
@@ -0,0 +1,74 @@
...
+        self.name = name
+        self.named_vars = {}
+        self.parent = model
+
+    @property
+    def model(self):
+        return self
+
+    @property
+    def decription(self):
```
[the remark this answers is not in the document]
sharanry: This is to give a summary of all RVs of the model along with their parameters. I will remove it for now?

pymc4/model.py
```diff
@@ -0,0 +1,74 @@
...
+        instance = super(Model, cls).__new__(cls)
+        if kwargs.get('model') is not None:
+            instance.parent = kwargs.get('model')
+        elif cls.get_contexts():
+            instance.parent = cls.get_contexts()[-1]
+        else:
+            instance.parent = None
+        return instance
+
+    def __init__(self, name="", model=None, ):
```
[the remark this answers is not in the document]
sharanry: This one way to enable nesting. Here `model=None` signifies that there is no parent model.

ColCarroll:
Note this was merged into a `pymc3_based_api` branch, in parallel with the `functional` branch

jolespin:
`PyMC4` is going to [use](https://coveralls.io/builds/17368946/source?filename=pymc4%2Fsample.py#L13) `xarray`  and `tensorflow`?  This is literally a dream come true!

Issue #2, opened 2018-05-30
Supported Python versions?


ColCarroll:
I will suggest developing on 3.6, but will add Python 3.5 to travis for now.  I would love to hear if there are good reasons to make sure to keep supporting 3.5, since 3.7 is coming out so soon.

leblancfg:
FWIW, Debian 9 (stable) uses Python 3.5, and has a projected end-of-life for August 2020.

ColCarroll:
Ah, thanks for that info!  f-strings and variable type annotations are probably [the features that we lose](https://docs.python.org/3/whatsnew/3.6.html) by supporting Python 3.5, but I don't think that's _so_ onerous.  I do love f-strings though...

junpenglao:
3.5 ignores the variable annotations right? so we can have them in the code anyway.

twiecki:
Agreed, 3.6 should be fine and we should make use of all new features we have.

ColCarroll:
Cool - I will update travis to start building on 3.6 and 3.7-dev!  Thanks for all the input!

PR #8, opened 2018-06-03
Test on 3.6 and 3.7-dev

ColCarroll:
Updating following the discussion in #2


ColCarroll:
Oh, I don't think we can test on 3.7-dev yet without being tricksy, because it can't install in Anaconda.

PR #9, opened 2018-06-13
Get mcmc sampling to work

sharanry:
- Unobserved variables accessible
- Sampling works
- log_prob_fn is given using the output of `f()`


sharanry:
@ferrine  
What do you suggest `model.target_log_prob_fn()` give? The logp function of only the result of `model.f` or all the intermediate RVs too?

If we give it only the final logp function then we wont be able to sample traces of intermediate RVs.
One more problem is `tfp.mcmc.sample_chain()` might work for only single RV(in current implementation,  final) logp function

pymc4/inference/sampling/sample.py
```diff
@@ -4,21 +4,21 @@
...
+           num_results=1000,
+           num_burnin_steps=500,
            step_size=.4,
            num_leapfrog_steps=3,
            numpy=True):
     initial_state = []
-    for name, shape in model.unobserved.iteritems():
-        initial_state.append(.5 * tf.ones(shape, name="init_{}".format(name)))
+    for name in model.unobserved:
+        initial_state.append(.5 * tf.ones(model.unobserved[name].shape, name="init_{}".format(name)))
```
ferrine: This will create a lot of problems with namespace

ferrine: I do not know the way to go but to avot repeated calls of `self._f`

sharanry:
``` python
unobserved = {}
        for i in self.variables:
            if self.variables[i] not in self.observed.values():
                unobserved[i] = self.variables[i]
        
        unobserved = collections.OrderedDict(unobserved)
        return unobserved
```

I could do this to avoid unobserved calling `f()` multiple times?

sharanry:
``` python
In [2]: graph = tf.Graph()
   ...:
   ...: with graph.as_default():
   ...:     ed.Normal(0., 1.)
   ...:     print(graph.as_graph_def())
   ...:
node {
  name: "Normal/loc/input"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
        }
        float_val: 0.0
      }
    }
  }
}
node {
  name: "Normal/loc"
  op: "Identity"
  input: "Normal/loc/input"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
}
node {
  name: "Normal/scale/input"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
        }
        float_val: 1.0
      }
    }
  }
}
node {
  name: "Normal/scale"
  op: "Identity"
  input: "Normal/scale/input"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
}
node {
  name: "Normal_1/sample/sample_shape"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
          }
        }
      }
    }
  }
}
node {
  name: "Normal_1/sample/Normal/batch_shape_tensor/batch_shape"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
          }
        }
      }
    }
  }
}
node {
  name: "Normal_1/sample/concat/values_0"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
            size: 1
          }
        }
        int_val: 1
      }
    }
  }
}
node {
  name: "Normal_1/sample/concat/axis"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
        }
        int_val: 0
      }
    }
  }
}
node {
  name: "Normal_1/sample/concat"
  op: "ConcatV2"
  input: "Normal_1/sample/concat/values_0"
  input: "Normal_1/sample/Normal/batch_shape_tensor/batch_shape"
  input: "Normal_1/sample/concat/axis"
  attr {
    key: "N"
    value {
      i: 2
    }
  }
  attr {
    key: "T"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "Tidx"
    value {
      type: DT_INT32
    }
  }
}
node {
  name: "Normal_1/sample/random_normal/mean"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
        }
        float_val: 0.0
      }
    }
  }
}
node {
  name: "Normal_1/sample/random_normal/stddev"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
        }
        float_val: 1.0
      }
    }
  }
}
node {
  name: "Normal_1/sample/random_normal/RandomStandardNormal"
  op: "RandomStandardNormal"
  input: "Normal_1/sample/concat"
  attr {
    key: "T"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "dtype"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "seed"
    value {
      i: 0
    }
  }
  attr {
    key: "seed2"
    value {
      i: 0
    }
  }
}
node {
  name: "Normal_1/sample/random_normal/mul"
  op: "Mul"
  input: "Normal_1/sample/random_normal/RandomStandardNormal"
  input: "Normal_1/sample/random_normal/stddev"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
}
node {
  name: "Normal_1/sample/random_normal"
  op: "Add"
  input: "Normal_1/sample/random_normal/mul"
  input: "Normal_1/sample/random_normal/mean"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
}
node {
  name: "Normal_1/sample/mul"
  op: "Mul"
  input: "Normal_1/sample/random_normal"
  input: "Normal/scale"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
}
node {
  name: "Normal_1/sample/add"
  op: "Add"
  input: "Normal_1/sample/mul"
  input: "Normal/loc"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
}
node {
  name: "Normal_1/sample/Shape"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
            size: 1
          }
        }
        int_val: 1
      }
    }
  }
}
node {
  name: "Normal_1/sample/strided_slice/stack"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
            size: 1
          }
        }
        int_val: 1
      }
    }
  }
}
node {
  name: "Normal_1/sample/strided_slice/stack_1"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
            size: 1
          }
        }
        int_val: 0
      }
    }
  }
}
node {
  name: "Normal_1/sample/strided_slice/stack_2"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
          dim {
            size: 1
          }
        }
        int_val: 1
      }
    }
  }
}
node {
  name: "Normal_1/sample/strided_slice"
  op: "StridedSlice"
  input: "Normal_1/sample/Shape"
  input: "Normal_1/sample/strided_slice/stack"
  input: "Normal_1/sample/strided_slice/stack_1"
  input: "Normal_1/sample/strided_slice/stack_2"
  attr {
    key: "Index"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "T"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "begin_mask"
    value {
      i: 0
    }
  }
  attr {
    key: "ellipsis_mask"
    value {
      i: 0
    }
  }
  attr {
    key: "end_mask"
    value {
      i: 1
    }
  }
  attr {
    key: "new_axis_mask"
    value {
      i: 0
    }
  }
  attr {
    key: "shrink_axis_mask"
    value {
      i: 0
    }
  }
}
node {
  name: "Normal_1/sample/concat_1/axis"
  op: "Const"
  attr {
    key: "dtype"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "value"
    value {
      tensor {
        dtype: DT_INT32
        tensor_shape {
        }
        int_val: 0
      }
    }
  }
}
node {
  name: "Normal_1/sample/concat_1"
  op: "ConcatV2"
  input: "Normal_1/sample/sample_shape"
  input: "Normal_1/sample/strided_slice"
  input: "Normal_1/sample/concat_1/axis"
  attr {
    key: "N"
    value {
      i: 2
    }
  }
  attr {
    key: "T"
    value {
      type: DT_INT32
    }
  }
  attr {
    key: "Tidx"
    value {
      type: DT_INT32
    }
  }
}
node {
  name: "Normal_1/sample/Reshape"
  op: "Reshape"
  input: "Normal_1/sample/add"
  input: "Normal_1/sample/concat_1"
  attr {
    key: "T"
    value {
      type: DT_FLOAT
    }
  }
  attr {
    key: "Tshape"
    value {
      type: DT_INT32
    }
  }
}
versions {
  producer: 26
}
```

For just one edward random variable there is some much change in the graph.

ferrine: That may be okay, if graph modifications are not frequently called (or it is hard to do) by user.

pymc4/model/base.py
```diff
@@ -68,14 +68,41 @@ def get_mode(state, rv, *args, **kwargs):
             returns = self.session.run(list(values_collector.result.values()))
         return dict(zip(values_collector.result.keys(), returns))
 
-    def target_log_prob_fn(self, *args, **kwargs):
+    def log_prob_fn(self, x,  *args, **kwargs):
```
ferrine: what is x for here?

pymc4/model/base.py
```diff
@@ -68,14 +68,41 @@ def get_mode(state, rv, *args, **kwargs):
...
 
+    def target_log_prob_fn(self, *args, **kwargs):
+        """
+        Unnormalized target density as a function of unobserved states.
+        """
+
+        def log_joint_fn(*args, **kwargs):
+            states = dict(zip(self.unobserved.keys(), args))
+            states.update(self.observed)
+            log_probs = []
```
ferrine: https://github.com/pymc-devs/pymc4/blob/functional/pymc4/util/interceptors.py#L110
```python
collect_log_prob = CollectLogProb(states)
with ed.interception(collect_log_prob):
    self._f(self._cfg)
return collect_log_prob.result
```

pymc4/model/base.py
```diff
@@ -67,6 +68,41 @@ def get_mode(state, rv, *args, **kwargs):
             returns = self.session.run(list(values_collector.result.values()))
         return dict(zip(values_collector.result.keys(), returns))
 
+    def log_prob_fn(self, x,  *args, **kwargs):
```
ferrine: hmm, I'm not sure this will work. ancestors of RV depend on the RV, here you do not replace RV with `value=kwargs.get(i)`

pymc4/model/base.py
sharanry: Not necessary, removing it.

pymc4/model/base.py
sharanry: changing it, was facing problems with states before. Now working.

sharanry: @ferrine `interceptors.CollectLogProb` only works with model like
``` python
@model.define
def process(cfg=None):
    mu = ed.Normal(0., 1., name="mu")
    obs = ed.Normal(0., 1., name="obs")
    return obs
```
and not model like
``` python
@model.define
def process(cfg=None):
    mu = ed.Normal(0., 1., name="mu")
    obs = ed.Normal(mu, 1., name="obs")
    return obs
```

ferrine: hmm, what's happening?

pymc4/inference/sampling/sample.py
```diff
@@ -10,7 +10,8 @@ def sample(model,
            num_leapfrog_steps=3,
            numpy=True):
     initial_state = []
-    for name, shape in model.unobserved.iteritems():
+    for name in model.unobserved.keys():
```
ColCarroll: use `for name, (_, shape, _) in model.items():` to indicate that `dist` and `rv` are not used in the loop

pymc4/model/base.py
```diff
@@ -83,6 +100,16 @@ def graph(self):
     def observed(self):
         return self._observed
 
+    @property
+    def unobserved(self):
+        unobserved = {}
+        for i in self.variables:
```
ColCarroll:
```python
for name, variable in self.variables.items():
    if variable not in self.observed.values():
        unobserved[name] = variable
```

pymc4/model/base.py
```diff
@@ -83,6 +100,16 @@ def graph(self):
     def observed(self):
         return self._observed
 
+    @property
+    def unobserved(self):
+        unobserved = {}
```
ColCarroll: this can be initialized as an `OrderedDict`: currently in Python < 3.6, the return value will not be ordered, since you built an (unordered) dictionary, then turned it into an `OrderedDict`.  We're targeting 3.6 and higher though, in which case you do not need the `OrderedDict` at all, since _all_ dicts now maintain insertion order.

That's a long way to say: I would just make a plain dictionary, but if you use `OrderedDict`, it needs to be initialized as such.

pymc4/util/interceptors.py
```diff
@@ -9,7 +9,7 @@
     'CollectLogProb'
 ]
 
-VariableDescription = collections.namedtuple('VariableDescription', 'Dist,shape')
+VariableDescription = collections.namedtuple('VariableDescription', 'Dist,shape,rv')
```
ferrine: I'm still worried about this solution. Variable description is supposed to be collected far before sampling (or what about changing this?). So when you first collect VariableInfo for the first time and get these RVs, you save temporary nodes. When you collect LogProb you again run the model and variables involves there are totally different from those that are stored in `VariableDescription`. That's why I did not store them there.

pymc4/model/base.py
ferrine: Nice point

pymc4/util/interceptors.py
sharanry: The problem i am facing if RVs are not stored in the VariableDescription is that it doesn't store the specifics of any distribution (like loc or scale) even if they are mentioned in the model definition. So we will have to collect all this already provided info somehow.

``` python
@model.define
def process():
    mu = ed.Normal(loc=0., scale=10., name="mu")
    # here we lose the info that it has loc 0 and scale 10 without RV.
```

We can try defining a new Interceptor which does this for us for each RV.
We can then overwrite(replace existing and add new) the collected data every-time we call configure.

sharanry:
@ferrine Could I merge this PR?

pymc4/model/base.py
```diff
@@ -67,6 +73,36 @@ def get_mode(state, rv, *args, **kwargs):
...
+        Pass the states of the RVs as args in alphabetical order of the RVs.
+        Compatible as `target_log_prob_fn` for tfp samplers.
+        """
+
+        def log_joint_fn(*args, **kwargs):  # pylint: disable=unused-argument
+            states = dict(zip(self.unobserved.keys(), args))
+            states.update(self.observed)
+            log_probs = []
+
+            def interceptor(f, *args, **kwargs):
```
ferrine: Don't we want use class based interceptor for consistency?

pymc4/util/interceptors.py
ferrine: OK, let's make it later

tests/test_model.py
```diff
@@ -49,3 +54,104 @@ def model(cfg):
...
+    @model.define
+    def simple(cfg):
+        ed.Normal(0., 1., name='normal')
+
+    model.observe(normal=1)
+
+    assert len(model.observed) == 1
+    assert not model.unobserved
+
+    model.reset()
```
ferrine: We decided to meke a copy of model each time state changes

ferrine: this one is not critical though, refactoring interceptor usage is what is really needed to finish this PR (https://github.com/pymc-devs/pymc4/pull/9#discussion-diff-200027548R87)

sharanry:
@ferrine I have changed it to a class based interceptor

ferrine:
And test point is better get via `model.test_point()`

sharanry:
@ferrine Currently `model.test_point()` is how you get the test point.
I am not sure I understand what you are saying.

pymc4/inference/sampling/sample.py
```diff
@@ -3,24 +3,24 @@
...
-def sample(model,
+def sample(model,  # pylint: disable-msg=too-many-arguments
            num_results=5000,
            num_burnin_steps=3000,
            step_size=.4,
            num_leapfrog_steps=3,
            numpy=True):
     initial_state = []
-    for name, shape in model.unobserved.iteritems():
+    for name, (_, shape, _) in model.unobserved.items():
```
ferrine:
```python
for name, point in model.test_point(mode=mode):
    initial_state.append(point)
```

ferrine: may be done in nest PR

PR #15, opened 2018-07-25
A simple solution to work on the copy of the graph

sharanry:


sharanry:
@ferrine, some problems with using the the right graph with `Session()`, that is why test is failing. Fixing it.

`self._cfg` contains the model configuration passed either during model initialization or using `model.configure()`

ColCarroll:
Looks nice - once tests are fixed, it might be worth wrapping this in its own decorator or method.

pymc4/model/base.py
```diff
@@ -83,9 +85,11 @@ def log_joint_fn(*args, **kwargs):  # pylint: disable=unused-argument
             states = dict(zip(self.unobserved.keys(), args))
             states.update(self.observed)
             interceptor = interceptors.CollectLogProb(states)
-            with ed.interception(interceptor):
+            temp_graph = tf.Graph()
+            tf.contrib.graph_editor.copy(self.graph, temp_graph)
+            with temp_graph.as_default(), ed.interception(interceptor):
                 self._f(self._cfg)
```
ColCarroll: not related to this PR, but this line should probably be `self._f(self.cfg)`, right?

sharanry: IMO it shouldnt make a difference. `self.cfg` returns `self._cfg`. Although I am not sure why we are saving it as `_cfg`

sharanry:
The CI is failing because they removed `tf-nightly==1.8.0.dev20180331`. Need to update the requirements.

pymc4/model/base.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, name=None, graph=None, session=None, **config):
             session = tf.Session(graph=graph)
         self.session = session
         self.observe(**config)
+        self.temp_graph = tf.Graph()
```
ColCarroll: I would use this pattern for creating and destroying the temporary graph.  I'm using a pymc3 model here because I think the pattern is similar, but let me know if it isn't!  Note that you will only need to implement the equivalent of the `temp_model` method below.

```python
import pymc3 as pm


def check_in_model_context():
    try:
        m = pm.modelcontext(None)
        return f'In model context! {len(m.vars)} vars!'
    except TypeError:
        return 'Not in model context!'

class Model(object):
    def __init__(self):
        self._temp_model = None
        
    @contextmanager
    def temp_model(self):
        self._temp_model = pm.Model()
        try:
            with self._temp_model:
                yield
        finally:
            self._temp_model = None
    
    def calculation(self):
        print(check_in_model_context())
        with self.temp_model():
            pm.Normal('x', 0, 1)
            print(check_in_model_context())
            
        print(check_in_model_context())
        with self.temp_model():
            print(check_in_model_context())
        print(check_in_model_context())
```

pymc4/util/interceptors.py
```diff
@@ -12,7 +12,7 @@
 VariableDescription = collections.namedtuple('VariableDescription', 'Dist,shape,rv')
 
 
-class Interceptor(object):
+class Interceptor():
```
ColCarroll: good practice to inherit from object, even if you don't need to!

twiecki: Why? I thought in py3 we don't need that anymore.

ferrine:
Explicit is better than implicit

вс, 29 июл. 2018 г., 15:03 Thomas Wiecki <notifications@github.com>:

pymc4/util/interceptors.py
sharanry: Yes, I am facing a linting error when I import from `object`

twiecki:
> Explicit is better than implicit

There is nothing explicit about this, it's a python 2 hack to get super working (or something like that).

pymc4/model/base.py
sharanry: decorator for inclass function is causing problems.. I will look into

pymc4/model/base.py
```diff
@@ -54,6 +55,7 @@ def _init_variables(self):
         info_collector = interceptors.CollectVariablesInfo()
         with self.graph.as_default(), ed.interception(info_collector):
             self._f(self.cfg)
+        tf.contrib.graph_editor.copy(self.graph, self.temp_graph)
```
ferrine: So 1) You create a temp graph for inner manipulations

pymc4/model/base.py
```diff
@@ -66,12 +68,15 @@ def not_observed(var, *args, **kwargs):  # pylint: disable=unused-argument
             def get_mode(state, rv, *args, **kwargs):  # pylint: disable=unused-argument
                 return rv.distribution.mode()
             chain.insert(0, interceptors.Generic(after=get_mode))
-
-        with self.graph.as_default(), ed.interception(interceptors.Chain(*chain)):
+        tf.contrib.graph_editor.copy(self.graph, self.temp_graph)
+        # pylint: disable=not-context-manager
+        with self.temp_graph.as_default(), ed.interception(interceptors.Chain(*chain)):
```
ferrine: You reuse tmp graph created on step 1) suppose `cfg` has *any* tensor from original graph and you call your model with `temp_graph.as_default()`. That creates all variables for `temp_graph`. This looks fine unless the case I'm talking about. If you attempt to do an operation on tow variables from different graphs, you will get an exception
image

ferrine: I do not know a solution for this case yet. I've seen some kind of working and relevant example in tensorflow codebase
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/function.py#L347

twiecki:
Closing as the main prototype is from the london branch.

PR #17, opened 2018-07-26
{WIP] Add HalfCauchy distribution

sharanry:


pymc4/distributions/half_cauchy.py
```diff
@@ -0,0 +1,160 @@
...
+  def _event_shape(self):
+    return tensor_shape.scalar()
+
+  def _sample_n(self, n, seed=None):
+    shape = tf.concat([[n], self.batch_shape_tensor()], 0)
+    probs = tf.random_uniform(
+        shape=shape, minval=0., maxval=1., dtype=self.dtype, seed=seed)
+    return tf.abs(sampled * self.scale)
+
+  def _prob(self, x):
```
twiecki: Don't we get all of that from the bijection applied to the Cauchy distribution?

sharanry: As @jvdillon mentioned in the tensorflow_probability [issue](https://github.com/tensorflow/probability/issues/100#issuecomment-405099938), the `AbsoluteValue` solution resembles HalfCauchy with similar properties but isn't exactly that. 

From what I understand, `AbsoluteValue` gives a modified scaled up version of Cauchy resulting in similar properties to HalfCauchy(with same scale as Cauchy described earlier) only on the positive side. This is demonstrated in this [notebook](https://gist.github.com/sharanry/e9552d3a74ec49e32e1e7fea4739e2b7). 

PS: The notebook uses HalfCauchy defined using absolute value.

IMO Only way to have a proper HalfCauchy distribution is to define it from scratch like tensorflow_probability does with [other distributions](https://github.com/tensorflow/probability/tree/master/tensorflow_probability/python/distributions).

twiecki: I see, that makes sense.

junpenglao:
Close as TFP now has HalfCauchy implemented.

PR #18, opened 2018-07-27
CI use pipenv as virtualenv and package manage tool

zyt312074545:
Install miniconda for each test not a good idea, It spend a lot of time, `pipenv` is a better tool to manage virtualenv and package. 
And use `pipenv` install package to avoid forgetting to write the package information into the `requirements.txt` or `requirements-dev.txt` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,6 @@ ENV/
 
 # VSCode
 .vscode/
+
+# Pycharm
+.idea/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,20 @@
+sudo: false
+
 language: python
 
 python:
   - "3.6"
 
+# command to install dependencies
 install:
-  - sudo apt-get update
-  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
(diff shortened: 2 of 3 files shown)


ColCarroll:
Thank you for the pull request! I'm hesitant to want to change the build script since `conda` supports the python scientific stack quite well, most developers on the project are used to working with it, and the current build time is under 3 minutes.  

The simplicity in `travis.yml` is super impressive, and I have been impressed with how intuitive `pipenv` is to use, I just don't see so many wins from using it here. I am closing this for now, but happy to reopen if I got something wrong!

zyt312074545:
@ColCarroll Thank for your reply! 

`pipenv` has a better feature of the `network` than other tools. If a error occurred while installing package, It will try again. Yes, `conda` is a good tool for scientific computing, but I don`t think CI need this feature. `pipenv` also can use `requirements.txt` as this dependencies file instead of `Pipfile`.

My bad, I don`t add environment variable `PIPENV_IGNORE_VIRTUALENVS` in `travis.yml` to cause the `python -m pytest -xv --cov=pymc4/ tests/` failed.

If you agree with my idea, I`m habby to modify the file and push to this branch.

PR #29, opened 2019-01-04
Store and use model arguments.

ColCarroll:
This tidies up model configuration. Here is a (forward) mixture model now. This example is in the notebook.

Note in particular that `mixture.configure` takes arguments that are passed to the model.

```python
import pymc4 as pm
import tensorflow as tf

@pm.model
def mixture(n_groups, n_points):    
    centers = pm.Normal('centers', loc=tf.zeros((n_groups, 2)), scale=1)
    scales = pm.HalfNormal('scales', 0.4 * tf.ones(n_groups))
    rates = pm.Dirichlet('rates', concentration=tf.ones(n_groups))
    
    group_assignments = pm.Multinomial('group_assignments', total_count=n_points, probs=rates)
    
    for idx in range(n_groups):
        count = tf.to_int32(group_assignments[idx])
        pm.Normal(f'group_{idx}', loc=centers[idx] * tf.ones((count, 2)), scale=scales[idx])
    
    
model = mixture.configure(n_groups=5, n_points=10_000)
```


ColCarroll:
Also, just made forward sampling return a dict instead of a list. I'd like this to be xarray eventually, but a dict is nice for now.

PR #31, opened 2019-01-06
Example Models

anhhuyalex:


twiecki:
Thanks! Can you make this as a separate NB?

anhhuyalex:
That should be done!

ericmjl:
@anhhuyalex yes, that would be awesome. The `pycache` files are auto-generated, so they should not be placed in source control.

.gitignore
```diff
@@ -1 +1,2 @@
-.ipynb_checkpoints/
\ No newline at end of file
+.ipynb_checkpoints/
+pymc4/__pycache__/
```
ColCarroll: needs a newline

canyon289:
All the `.pyc` files need to be removed from the gitversion tree I believe, or git will never ignore them

Could use a second consensus on this one

canyon289:
Before we merge this please squash the git tree so the `pycache` files don't enter the git graph.

anhhuyalex:
@canyon289 Could you clarify what you mean by squashing this branch? I merged code from pymc/master to my branch to work with the latest version of the code. Would squashing be a problem then?

canyon289:
Ah sorry for not explaining
Squashing means taking 2 or more commits, and "squashing" them together into one commit. This picture is the best visual explanation I could find.

image

The reason I ask is that we want to make sure the __pycache__ files are not in the git tree. Right now since they were deleted in two commits, git tracks those files. If all the commits are squashed together, the two actions will cancel out.

Merging your code form master is not a problem! Things should still work out.

Lastly I can offer to do this for you when you're done with the PR, the one caveat is that it will look like the pull request came from me due to the way git works. If you're ok with that I don't mind doing it.

If you aren't I would be happy to type out the commands to get you going.

Let me know, and above all thank you very much for this contribution! We really appreciate it

canyon289:
@anhhuyalex Let me know if you'd still like help, or if you want me to squash and force push here! Happy to get you unblocked

anhhuyalex:
@canyon289 Sorry for the late reply. I was trying to figure out how git works, since I'm a beginner to open source. :)

I tried what you suggested, so let me know if there's anything else I can do.

canyon289:
Yes! This is perfect! Thank you. No apologies needed. Git is extremely confusing and there's no shame in being confused by it. I'm still confused by git regularly.

I'll do a final review today but things are looking great. Much appreciated

canyon289:
Thanks for the contribution

Issue #32, opened 2019-01-09
Need requirements.txt


sharanry:
Would be helpful for people to contribute with the addition of `requirements.txt`.

eigenfoo:
Closed by #33

PR #33, opened 2019-01-09
Add requirements.txt

sharanry:
#32 

Is it necessary to add pymc3 and theano? They are being used in `_hmc`.
Will `pymc3==3.6` and `Theano==1.0.1` do?


twiecki:
Great! I don't think that code is functional yet and once it is will probably get detached so we can leave it for now.

junpenglao:
I think it would be better to use tf-nightly and tfp-nightly

PR #34, opened 2019-01-11
Change evaluate return value for clarity

canyon289:
Add gitignore

### DIFF
--- a/pymc4/_model.py
+++ b/pymc4/_model.py
@@ -32,7 +32,8 @@ class Model:
     def _evaluate(self):
         """Call the template function with the saved arguments."""
         args, kwargs = self._template_args
-        return self._template._func(*args, **kwargs)
+        self._template._func(*args, **kwargs)
+        return
 
     def make_log_prob_function(self):
         def log_prob(*args):


canyon289:
Removed gitignore since its in the other PR

pymc4/_model.py
```diff
@@ -32,7 +32,8 @@ def __init__(self, template, forward_context, template_args):
     def _evaluate(self):
         """Call the template function with the saved arguments."""
         args, kwargs = self._template_args
-        return self._template._func(*args, **kwargs)
+        self._template._func(*args, **kwargs)
+        return
```
twiecki: I think just get rid of the return, looks like a bug to me otherwise.

canyon289: Interesting, I usually prefer the explicit return, even if its none, because it helps me visually bookmark the end of functions.

If this is a strong preference I can do it in another PR

ericmjl: Never thought of a blank "return" as being a "bookmarker", @canyon289. I learned something new today! That said, I think it isn't Pythonic convention to put a "blank return" the way you have. Is there something in the Zen of Python we can refer to?

canyon289: I don't think there's any strong reference I can point to than personal preference.

That being said my personal preference is not a great argument to keep it!

PR #35, opened 2019-01-11
Add names to Tensorflow graph distributions

canyon289:
Adding the name to the distribution makes it much simpler to debug the tensorflow graph and understand which tensorflow nodes are associated with which pm.var

### DIFF
--- a/SimpleModels.ipynb
+++ b/SimpleModels.ipynb
@@ -13,7 +13,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import pymc4 as pm\n",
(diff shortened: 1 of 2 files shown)


canyon289:
Before: https://github.com/pymc-devs/pymc4/blob/91aac796d771072dd4ef8cce06761d7b08798a4d/SimpleModels.ipynb

After:
https://github.com/pymc-devs/pymc4/blob/6883e0790fae889eac3e61680e576817ac48b037/SimpleModels.ipynb

ColCarroll:
Neat! LGTM.

PR #36, opened 2019-01-11
Add setup file for package portability

canyon289:
Makes it easier to develop if we don't have to worry where the notebooks are relative to the source code


ColCarroll:
TODO: 
image

canyon289:
I'm not sure what that should be but figured that would be (humorously) fine for now

PR #41, opened 2019-01-12
Wrap tfp distributions as pm RVs

eigenfoo:
(Following up @twiecki's suggestion)

This PR wraps `tfp.distribution`s as `pm.RandomVariable`s. We enter the list of tfp distributions we want as strings in `__all__`, and the for loop at the end of the module should define the necessary classes.

Some notes:
- I assumed that we would name our RVs the exact same name as the tfp distributions (e.g., it's now `Chi2`, not `ChiSquared`).
- I made the list of distributions by just comparing what RVs we have in PyMC3 right now. I'll double check this list later today.


ColCarroll:
Cool! I guess since you set `__all__`, this allows tab completion?

I might use 

```
setattr(sys.modules[__name__], dist_name, getattr(tfp.distributions, dist_name))
```

because I like `setattr` and `getattr` better. Would that even work? [This suggests they are similar](https://stackoverflow.com/questions/2933470/how-do-i-call-setattr-on-the-current-module), and gives no prohibition against either. I guess take this as an observation that there's another way to do it, then! 

Looks good to merge, but I'll give it a few more hours in case there's input.

eigenfoo:
@ColCarroll That does indeed work! I like `getattr` and `setattr` better: I didn't know that they worked at the module level. I've changed it to do that.

And no, I manually typed out the distributions, since I use vim with very few plugins 😛

eigenfoo:
Looking carefully, tfp does not implement the following distributions, which are currently supported in PyMC3. There are some fairly important ones here: we'll need to decide what we want to do there (i.e. implement the RVs ourselves, submit PRs to tfp, or drop support for the obscure distributions).

Continuous:
- Flat
- HalfFlat
- SkewNormal
- HalfStudentT
- Weibull
- Wald
- ExGaussian
- LogitNormal
- Interpolated

Discrete:
- BetaBinomial
- ZeroInflatedBinomial
- ZeroInflatedPoisson
- ZeroInflatedNegativeBinomial
- DiscreteUniform
- Constant
- DiscreteWeibull
- OrderedLogistic

Multivariate:
- MatrixNormal
- KroneckerNormal
- LKJCholeskyCov and LKJCorr

Timeseries:
- All Timeseries distributions

ColCarroll:
This will be nice, not adding them piecemeal. Thanks, @eigenfoo!

majidaldo:
What happens if there is a distribution that can't be based of a (existing) TFP distribution? TFD Distributions have so many methods. In PyMC3 you could at least attempt to provide your own .log_p() and sample().

PR #42, opened 2019-01-12
Add travis.yml for black check

canyon289:
I don't have permissions to setup anything on the TravisCI side, or in this repo, but would be happy to do so if I them!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ install:
 
 script:
   - black pymc4/ --check
+
+  # Add these in for now so we can see what version was available at time CI check
+  - pip install -r requirements.txt
   - python -c "import tensorflow; print(tensorflow.__version__)"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+python:
+  - "3.6"
+sudo: required
+
+install:
+  - pip install black
+  -
+
+script:
+  - black pymc4/ --check
```
eigenfoo: We should set the `--py36` flag.

Also, did we agree on line lengths of 100 or something? I'm not sure.

canyon289: I don't think we explicitly agreed. In the other PR I set 100 because that's what we did in ArviZ and it got merged so ¯\_(ツ)_/¯

Good idea on py36 flag

canyon289: I'll do the 3.6 flag in another PR since it will change the formatting

canyon289:
Ready for review

PR #45, opened 2019-01-13
Add .gitignore

eigenfoo:


ericmjl:
This is great! Thanks for doing this, @eigenfoo!

Issue #44, opened 2019-01-12
Support more distributions as random variables


eigenfoo:
Filing this as an issue so we don't forget: https://github.com/pymc-devs/pymc4/pull/41#issuecomment-453775302

I think we can achieve most of the important distributions using `tfp.distributions.Mixture` and `tfp.distributions.Bijector` (e.g. the zero-inflated distributions are just mixtures with the zero distribution).

@ColCarroll is there a concrete problem with having a distribution/RV named `Deterministic`? That's what tfp calls the thing that PyMC calls the `Constant` distribution. If so, I can write something to remap names

junpenglao:
LKJ in tfp is the same as LKJCorr.

eigenfoo:
Edited!

eigenfoo:
Update: #46 merges in some more random variables (those that can be created via tfp's bijectors and transformed variables). There are still quite a few random variables outstanding, though.

ericmjl:
@eigenfoo your PR gives me a good template to work off for some distributions. I'll take a stab at it this afternoon!

ericmjl:
@eigenfoo your PR gives me a template to work off! I'll see how fast I can get setup and running for development at work.

junpenglao:
FYI: LKJ in tfp now has Cholesky parameterization https://github.com/tensorflow/probability/commit/9b007d8e52074fe973cf06290cf5411274636834

tblazina:
I haven't contributed yet but would be interested to have a go with helping with some of these. Looking at the code base it seems a lot of these still are not implemented, so it is still open, correct?

ericmjl:
@tblazina yes, still open! I recently learned a new pattern at this year's PyCon, which helps with tracking progress. Was wondering what the other devs think of this - when putting in a PR, put in also an issue that we can close, and reference it in the branch name (e.g. `distribution-flat-#93`)? I think there's github magic that we can take advantage of :smile:

tblazina:
@ericmjl, ok great, thanks for quick response! I'll try to get dev environment all set up and see if I can get plugging away at some of this. I think your idea about tracking progress sounds to me and was along the lines of my next question, is that for a big issue like this with lots of different parts, is the best practice in this project to submit a PR for each distribution that we implement and simply reference this issue in the PR?

ericmjl:
> is the best practice in this project to submit a PR for each distribution that we implement and simply reference this issue in the PR?

[...]

PR #46, opened 2019-01-13
ENH: add more random variables

eigenfoo:
First attempt at #44. Not ready for review yet. This PR should wait until after #41 is merged.

Briefly, there are two kinds of random variables: 1) those that are supported as tfp distributions, and 2) those that are not. Type 1 is easily wrapped as RVs. A good number of type 2 can be achieved by `tfd.TransformedDistribution`, `tfd.Bijector` and `tfd.Mixture`. I hope that this PR will merge in all the distributions that PyMC3 currently supports, that can also be made from mixtures or bijectors of type 1 distributions. This will leave us with the trickier distributions to implement later (e.g. SkewNormal).

Some other notes:
- There should be a restructuring PR following this to split out all this code into several files.
- This PR renames PyMC3's `Wald` to tfp's `InverseGaussian`

### DIFF
--- a/pymc4/_random_variables.py
+++ b/pymc4/_random_variables.py
@@ -12,8 +12,8 @@ import tensorflow_probability as tfp
 from tensorflow_probability import distributions as tfd
 
 
-# Random variables that PyMC3 supports, but tfp does not support as
-# distributions. We implement these random variables.
+# Random variables tfp does not support as distributions. We implement these
+# random variables.
 tfp_unsupported = [
     "Constant",
     "DiscreteUniform",
@@ -25,9 +25,9 @@ tfp_unsupported = [
     "ZeroInflatedPoisson",
 ]
 
-# Random variables that PyMC3 supports, that tfp also support as distributions.
-# We wrap these distributions as random variables. Names must match
-# tfp.distributions names exactly.
+# Random variables that tfp supports as distributions. We wrap these
+# distributions as random variables. Names must match tfp.distributions names
+# exactly.
 tfp_supported = [
     "Bernoulli",
     "Beta",
--- a/pymc4/_random_variables.py
+++ b/pymc4/_random_variables.py
@@ -212,7 +212,7 @@ class Constant(RandomVariable):
 
 
 class DiscreteUniform(RandomVariable):
-    def __dist(*args, **kwargs):
+    def _base_dist(*args, **kwargs):
         probs = np.ones(high - low) / (high - low)
         return tfd.TransformedDistribution(
             distribution=tfd.Categorical(probs=probs),
@@ -220,83 +220,69 @@ class DiscreteUniform(RandomVariable):
             name="DiscreteUniform",
         )
 
-    _base_dist = __dist
-
 
 class HalfStudentT(RandomVariable):
     # A HalfStudentT is the absolute value of a StudentT.
-    def __dist(*args, **kwargs):
+    def _base_dist(*args, **kwargs):
         return tfd.TransformedDistribution(
             distribution=tfd.StudentT(*args, **kwargs),
             bijector=tfp.bijectors.AbsoluteValue(),
             name="HalfStudentT",
         )
 
-    _base_dist = __dist
-
 
 class LogitNormal(RandomVariable):
     # A LogitNormal is the standard logistic of a Normal.
-    def __dist(*args, **kwargs):
+    def _base_dist(*args, **kwargs):
         return tfd.TransformedDistribution(
             distribution=tfd.Normal(*args, **kwargs),
             bijector=tfp.bijectors.Sigmoid(),
             name="LogitNormal",
         )
 
-    _base_dist = __dist
-
 
 class Weibull(RandomVariable):
     # The inverse of the Weibull bijector applied to a U[0, 1] random variable
     # gives a Weibull-distributed random variable.
-    def __dist(*args, **kwargs):
+    def _base_dist(*args, **kwargs):
         return tfd.TransformedDistribution(
             distribution=tfd.Uniform(0.0, 1.0),
             bijector=tfp.bijectors.Invert(tfp.bijectors.Weibull(*args, **kwargs)),
             name="Weibull",
         )
 
-    _base_dist = __dist
-
 
 class ZeroInflatedBinomial(RandomVariable):
     # A ZeroInflatedBinomial is a mixture between a deterministic distribution
     # and a Binomial distribution.
-    def __dist(mix, *args, **kwargs):
+    def _base_dist(mix, *args, **kwargs):
         return tfd.Mixture(
             cat=tfd.Categorical(probs=[mix, 1.0 - mix]),
             components=[tfd.Deterministic(0.0), tfd.Binomial(*args, **kwargs)],
             name="ZeroInflatedBinomial",
         )
 
-    _base_dist = __dist
-
 
 class ZeroInflatedPoisson(RandomVariable):
     # A ZeroInflatedPoisson is a mixture between a deterministic distribution
     # and a Poisson distribution.
-    def __dist(mix, *args, **kwargs):
+    def _base_dist(mix, *args, **kwargs):
         return tfd.Mixture(
             cat=tfd.Categorical(probs=[mix, 1.0 - mix]),
             components=[tfd.Deterministic(0.0), tfd.Poisson(*args, **kwargs)],
             name="ZeroInflatedPoisson",
         )
 
-    _base_dist = __dist
-
 
 class ZeroInflatedNegativeBinomial(RandomVariable):
     # A ZeroInflatedNegativeBinomial is a mixture between a deterministic
     # distribution and a NegativeBinomial distribution.
-    def __dist(mix, *args, **kwargs):
+    def _base_dist(mix, *args, **kwargs):
         return tfd.Mixture(
             cat=tfd.Categorical(probs=[mix, 1.0 - mix]),
             components=[tfd.Deterministic(0.0), tfd.NegativeBinomial(*args, **kwargs)],
             name="ZeroInflatedNegativeBinomial",
         )
-
-    _base_dist = __dist
 
 
 # Programmatically wrap tfp.distribtions into pm.RandomVariables
--- a/pymc4/_random_variables.py
+++ b/pymc4/_random_variables.py
@@ -217,7 +217,7 @@ class DiscreteUniform(RandomVariable):
         kwargs.update({"low": low, "high": high})
         super(DiscreteUniform, self).__init__(name, *args, **kwargs)
 
-    def _base_dist(*args, **kwargs):
+    def _base_dist(self, *args, **kwargs):
         """A DiscreteUniform is an equiprobable Categorical over (high-low),
         shifted up by low."""
         low = kwargs.pop("low")
@@ -231,7 +231,7 @@ class DiscreteUniform(RandomVariable):
 
 
 class HalfStudentT(RandomVariable):
-    def _base_dist(*args, **kwargs):
+    def _base_dist(self, *args, **kwargs):
         """A HalfStudentT is the absolute value of a StudentT."""
         return tfd.TransformedDistribution(
             distribution=tfd.StudentT(*args, **kwargs),
@@ -241,7 +241,7 @@ class HalfStudentT(RandomVariable):
 
 
 class LogitNormal(RandomVariable):
-    def _base_dist(*args, **kwargs):
+    def _base_dist(self, *args, **kwargs):
         """A LogitNormal is the standard logistic of a Normal."""
         return tfd.TransformedDistribution(
             distribution=tfd.Normal(*args, **kwargs),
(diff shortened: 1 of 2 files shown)


pymc4/_random_variables.py
```diff
@@ -1,12 +1,75 @@
...
-import tensorflow as tf
+import sys
+import numpy as np
 import tensorflow_probability as tfp
-
-__all__ = ["Normal", "HalfNormal", "Multinomial", "Dirichlet"]
+from tensorflow_probability import distributions as tfd
+
+
+# Random variables that PyMC4 support, but tfp does not support as
```
fonnesbeck: You mean PyMC3? PyMC4 currently doesn't have any of these.

eigenfoo: Oops sorry, typo!

pymc4/_random_variables.py
```diff
@@ -7,11 +7,28 @@
...
 
 import sys
-import tensorflow_probability.distributions as tfd
-
+import numpy as np
+import tensorflow_probability as tfp
+from tensorflow_probability import distributions as tfd
+
+
+# Random variables that PyMC3 support, but tfp does not support as
```
twiecki: PyMC4? :)

pymc4/_random_variables.py
```diff
@@ -184,8 +204,103 @@ def as_tensor(self):
...
+            bijector=tfp.bijectors.AffineScalar(shift=low),
+            name="DiscreteUniform",
+        )
+
+    _base_dist = __dist
+
+
+class HalfStudentT(RandomVariable):
+    # A HalfStudentT is the absolute value of a StudentT.
+    def __dist(*args, **kwargs):
```
twiecki: Why not call this `_base_dist` directly?

pymc4/_random_variables.py
eigenfoo: @fonnesbeck was confused when I originally wrote PyMC4! 😅

PyMC4 _currently_ doesn't have any of those random variables, but after this PR it will... so I think we should be saying PyMC4?

twiecki: Oh now I get it, it's fine then.

twiecki: *supports

twiecki:
Looks great! Still WIP?

eigenfoo:
Status update & several worrying things:

- AFAICT all of the `tfp_supported` random variables work: you can instantiate them, sample from them, and compute their `log_prob`. Yay!
- All of the `tfp_unsupported` random variables (asides from `DiscreteUniform`, for some reason) cannot even be instantiated: we run into context errors (reproducible example and stack trace below). I don't really know what to do here 😄

A lot of these thorns will get ironed out with a test suite, which I'll work on soon, but I'm not sure if we want to merge this since we have context errors.

<details>
  <summary>Click to expand</summary>
  <p>

```
In [10]: pm.HalfStudentT(name='x', loc=0, scale=1)
AttributeError                            Traceback (most recent call last)
<ipython-input-10-4ed65ed7b308> in <module>()
----> 1 pm.HalfStudentT(name='x', loc=0, scale=1)

~/Desktop/pymc4/pymc4/_random_variables.py in __init__(self, name, *args, **kwargs)
    180     def __init__(self, name, *args, **kwargs):
    181         self._parents = []
--> 182         self._distribution = self._base_dist(name=name, *args, **kwargs)
    183         self._sample_shape = ()
    184         self._dim_names = ()

~/Desktop/pymc4/pymc4/_random_variables.py in _base_dist(*args, **kwargs)
    235         """A HalfStudentT is the absolute value of a StudentT."""
    236         return tfd.TransformedDistribution(
--> 237             distribution=tfd.StudentT(*args, **kwargs),
    238             bijector=tfp.bijectors.AbsoluteValue(),
    239             name="HalfStudentT",

~/miniconda3/lib/python3.6/site-packages/tensorflow_probability/python/distributions/student_t.py in __init__(self, df, loc, scale, validate_args, allow_nan_stats, name)
    172     with tf.name_scope(name, values=[df, loc, scale]) as name:
    173       dtype = dtype_util.common_dtype([df, loc, scale], tf.float32)
--> 174       df = tf.convert_to_tensor(df, name="df", dtype=dtype)
    175       loc = tf.convert_to_tensor(loc, name="loc", dtype=dtype)
    176       scale = tf.convert_to_tensor(scale, name="scale", dtype=dtype)

~/miniconda3/lib/python3.6/site-packages/tensorflow/python/framework/ops.py in convert_to_tensor(value, dtype, name, preferred_dtype)
   1048       name=name,
   1049       preferred_dtype=preferred_dtype,
-> 1050       as_ref=False)
   1051
   1052

~/miniconda3/lib/python3.6/site-packages/tensorflow/python/framework/ops.py in internal_convert_to_tensor(value, dtype, name, as_ref, preferred_dtype, ctx)
   1144
   1145     if ret is None:
-> 1146       ret = conversion_func(value, dtype=dtype, name=name, as_ref=as_ref)
   1147
   1148     if ret is NotImplemented:

~/Desktop/pymc4/pymc4/__init__.py in _convert_rv_to_backend(d, dtype, name, as_ref)
     12 def _convert_rv_to_backend(d, dtype=None, name=None, as_ref=False):
     13     if isinstance(d, _RandomVariable):
---> 14         return d.as_tensor()
     15     return d
     16

~/Desktop/pymc4/pymc4/_random_variables.py in as_tensor(self)
    197     def as_tensor(self):
    198         ctx = contexts.get_context()
--> 199         if id(ctx) != self._creation_context_id:
    200             raise ValueError("Cannot convert to tensor under new context.")
    201         if self._backend_tensor is None:

AttributeError: 'HalfStudentT' object has no attribute '_creation_context_id'
```

  </p></details>
</p></details>

twiecki:
Maybe @csuter has an idea?

eigenfoo:
Is this really an upstream tfp issue? It looks like it's something to do with the model context manager, I'm just not sure because I'm unfamiliar with the context management parts of PyMC4.

csuter:
I suspect there is a problem with the way you're forwarding `name`, `*args` and `**kwargs` to _base_dist(). Downstream, somehow the thing getting passed as `df` to `tfd.StudentT` is an object which has been registered for conversion to tensor by PyMC code; it looks like it is some RandomVariable instance. I'd start by 1. inspecting the *args and **kwargs variables, in when base_dist is called, 2. (possibly) pushing "name=..." to the end of the args list (instead of before *args and **kwargs)..I'm not sure how the mixing of non-varargs and varargs works exactly, but I wouldn't be surprised if what's written is doing something non-obvious.

csuter:
Oh now I see it -- you're calling self._base_dist but you haven't declared a self param in your implementation of _base_dist. So the first entry in *args is self (as defined in the constructor). So you're passing on a not-yet-fully-constructed thing, ultimately into tf.convert_to_tensor, which looks at the context manager stuff and finds that it's not set up yet (probably because construction isn't complete!).

varargs are tricky! ;)

eigenfoo:
Ah I see, thank you @csuter! Very insightful. varargs _are_ tricky! I've added a `self` param to all the `_base_dist`s now.

Unfortunately, running `import pymc4 as pm` now raises an error: `AttributeError: module 'tensorflow' has no attribute 'distributions'`. It looks like this is an upstream bug in tfp?

<details>
  <summary>Click to expand</summary>
  <p>

```
In [1]: import pymc4 as pm
AttributeError                            Traceback (most recent call last)
<ipython-input-1-3498ab11c487> in <module>
----> 1 import pymc4 as pm

~/Desktop/pymc4/pymc4/__init__.py in <module>
      1 """PyMC4"""
      2 from ._model import *
----> 3 from ._random_variables import *
      4
      5 import tensorflow as _tf

~/Desktop/pymc4/pymc4/_random_variables.py in <module>
      9 import sys
     10 import numpy as np
---> 11 import tensorflow_probability as tfp
     12 from tensorflow_probability import distributions as tfd
     13

~/Desktop/pymc4/venv/lib/python3.6/site-packages/tensorflow_probability/__init__.py in <module>
     76
     77 # from tensorflow_probability.google import staging  # DisableOnExport
---> 78 from tensorflow_probability.python import *  # pylint: disable=wildcard-import
     79 from tensorflow_probability.python.version import __version__
     80 # pylint: enable=g-import-not-at-top

~/Desktop/pymc4/venv/lib/python3.6/site-packages/tensorflow_probability/python/__init__.py in <module>
     20
     21 from tensorflow_probability.python import bijectors
---> 22 from tensorflow_probability.python import distributions
     23 from tensorflow_probability.python import edward2
     24 from tensorflow_probability.python import glm

~/Desktop/pymc4/venv/lib/python3.6/site-packages/tensorflow_probability/python/distributions/__init__.py in <module>
     22 from tensorflow_probability.python.distributions.autoregressive import Autoregressive
     23 from tensorflow_probability.python.distributions.batch_reshape import BatchReshape
---> 24 from tensorflow_probability.python.distributions.bernoulli import Bernoulli
     25 from tensorflow_probability.python.distributions.beta import Beta
     26 from tensorflow_probability.python.distributions.binomial import Binomial

~/Desktop/pymc4/venv/lib/python3.6/site-packages/tensorflow_probability/python/distributions/bernoulli.py in <module>
    153
    154 # TODO(b/117098119): Remove tf.distribution references once they're gone.
--> 155 @kullback_leibler.RegisterKL(Bernoulli, tf.distributions.Bernoulli)
    156 @kullback_leibler.RegisterKL(tf.distributions.Bernoulli, Bernoulli)
    157 @kullback_leibler.RegisterKL(Bernoulli, Bernoulli)

AttributeError: module 'tensorflow' has no attribute 'distributions'
```

  </p></details>
</p></details>

twiecki:
We should also add a test that goes through the list of all RVs, instantiates them, draws values from them etc. We don't have to test specific values (we rely on TFP for correctness), just that it doesn't raise exceptions.

canyon289:
> Ah I see, thank you @csuter! Very insightful. varargs _are_ tricky! I've added a `self` param to all the `_base_dist`s now.
> 
> Unfortunately, running `import pymc4 as pm` now raises an error: `AttributeError: module 'tensorflow' has no attribute 'distributions'`. It looks like this is an upstream bug in tfp, which seems bizarre to me since I didn't have this problem yesterday.
> 
> Click to expand

The namespacing changed in TFP nightly. Inspecting the source code shows the change.
Try `from tfp import distributions as tfd`

canyon289:
> We should also add a test that goes through the list of all RVs, instantiates them, draws values from them etc. We don't have to test specific values (we rely on TFP for correctness), just that it doesn't raise exceptions.

I'm doing this in the other PR already :)

csuter:
> > Ah I see, thank you @csuter! Very insightful. varargs _are_ tricky! I've added a `self` param to all the `_base_dist`s now.
> > Unfortunately, running `import pymc4 as pm` now raises an error: `AttributeError: module 'tensorflow' has no attribute 'distributions'`. It looks like this is an upstream bug in tfp, which seems bizarre to me since I didn't have this problem yesterday.
> > Click to expand
> 
> The namespacing changed in TFP nightly. Inspecting the source code shows the change.
> Try `from tensorflow_probability import distributions as tfd`

Right. If you're using tf-nightly, you should use tfp-nightly. If you're using tensorflow (stable) you should use the corresponding stable tensorflow-probability release (see release notes on our GH page for compatibility info). We don't explicitly depend on TF in our setuptools config because users have to choose between GPU and non-GPU versions of TF (not ideal; this will change in the somewhat near term). Until then, we (TFP) raise an exception when users try to use incompatible (stable) TF/TFP versions. We don't have a proper check in place for nightly; we need to add it...

eigenfoo:
@twiecki ready for final review and merge. Doubtless this code has a few bugs in it, but we're going to catch them as tests get added (hopefully soon!)

pymc4/_random_variables.py
```diff
@@ -44,6 +62,8 @@
     "Wishart",
 ]
 
+__all__ = tfp_supported + tfp_unsupported
```
canyon289: If we do this how will the user know which distributions they can put into a tensorflow graph and which ones they can't?

canyon289: Or do they all work because of the additional code below?

canyon289:
> @twiecki ready for final review and merge. Doubtless this code has a few bugs in it, but we're going to catch them as tests get added (hopefully soon!)

Do you want to wait for testing then before merging?

pymc4/_random_variables.py
twiecki: They all work, I think that comment about pymc3 is a bit confusing, even though it makes sense once you know what is meant. @eigenfoo maybe we can reword that or just get rid of it.

canyon289:
It might take 2 to 5 days for me to get the entire CI infrastructure going with tests.

If that's too long maybe you can merge and we'll retroactively fix?

eigenfoo:
I'm not sure how much time I can guarantee to work on this PR in a week's time, so I'd prefer to merge this sooner. If it turns out that I can dedicate more time to this, I'd be happy to go back and fix it. I don't like to break master, but from what I understand master is already a WIP that doesn't work.

twiecki:
@eigenfoo so it's working (you tested it locally) but there are no tests?

eigenfoo:
Correct. I checked that a few random variables could sample, logp, etc. by hand. Everything I tried worked.

twiecki:
Awesome, thanks @eigenfoo, this moves us ahead a big step!

canyon289:
Thanks @eigenfoo!

PR #49, opened 2019-01-14
Local Pytest and docker TravisCI testing

canyon289:
Adding the following in this PR. Will unmark WIP when asking for a merge
* More test infrastructure (In addition and non conflicting with https://github.com/pymc-devs/pymc4/pull/39)
* pylint
* requirements-dev.txt
* TravisCI testing, and linting
* Coveralls


canyon289:
This is the issue we were having on ArviZ with tensorflow, which is why we ended up moving to docker testing. Seems like its a problem here too

https://travis-ci.org/pymc-devs/pymc4/builds/479554525#L609

requirements.txt
```diff
@@ -1,2 +1,4 @@
-tf-nightly-2.0-preview
+tf-nightly
```
twiecki: why change this back?

canyon289: TFP-nightly and tf-nightly don't seem to work together. tfp-nightly trys to import tf.contrib which from my understanding, and what the importerror tells me, not available anymore.

canyon289:
Well at this point tests and lint are working if anyone wants to take a look.

I need to do a few more things to button up this pull request, along the lines of add documentation and propose a couple of things. Please feel free to review in the meanwhile though

pymc4/tests/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+PyMC4 test configuration
+"""
+import pytest
+import tensorflow as tf
+
+
+@pytest.fixture(scope="function")
+def tf_session(request):
+    sess = tf.Session()
+
```
twiecki: or could just `yield sess` and then `reset_default_graph()`.

scripts/install_miniconda.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
twiecki: Why is this required if the docker image pulls from the miniconda image?

canyon289:
This is ready for review

kyleabeauchamp:
What conda channels do we plan on using?  It appears the conda docker is only pulling from defaults, but we may want to grab things (tensorflow?) from conda-forge.

kyleabeauchamp:
I imagine we should test against a similar conda channel setup that we eventually want users to adopt...

canyon289:
> What conda channels do we plan on using? It appears the conda docker is only pulling from defaults, but we may want to grab things (tensorflow?) from conda-forge.

Fair point, right now we had some discussions about using `tf-nightly` and `tfp-nightly` and then switching to tf2 when tfp becomes compatible.

Besides those to requirements though nothing is installed by conda specifically, other than the base env packages. Everything else is done by pip

kyleabeauchamp:
If we eventually want to distribute on `conda-forge` then users will likely be using a python binary from conda-forge (https://anaconda.org/conda-forge/python/files?version=3.6.7) rather than defaults.  Regardless we can sort this out later, just wanted to chime in.

canyon289:
> If we eventually want to distribute on `conda-forge` then users will likely be using a python binary from conda-forge (https://anaconda.org/conda-forge/python/files?version=3.6.7) rather than defaults. Regardless we can sort this out later, just wanted to chime in.

I certainly agree with you on this. Not sure where we should record this thought, or if these comments on this PR are enough

pymc4/tests/test_random_variables.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_tf_session_cleared(tf_session):
+    """Temporary test: Check that fixture is finalizing correctly"""
+    assert len(tf_session.graph.get_operations()) == 0
+
+
+@pytest.mark.parametrize(
+    "tf_distribution", _random_variables.tfp_supported, ids=_random_variables.tfp_supported
+)
+def test_tfp_distributions(tf_session, tf_supported_args, tf_distribution):
```
twiecki: these are great!

pymc4/tests/test_random_variables.py
```diff
@@ -0,0 +1,97 @@
...
+    if tf_distribution is not "Binomial":
+        # Assert that values are returned with no exceptions
+        log_prob = dist.log_prob()
+        vals = tf_session.run([log_prob], feed_dict={dist._backend_tensor: sample})
+        assert vals is not None
+
+    else:
+
+        # Bionomial distribution raises exception when calling log_prob
+        assert tf_distribution == "Binomial"
```
twiecki: this seems superfluous :)

pymc4/tests/test_random_variables.py
```diff
@@ -0,0 +1,97 @@
...
+def test_tfp_distributions(tf_session, tf_supported_args, tf_distribution):
+    """Test all TFP supported distributions"""
+    _dist = getattr(_random_variables, tf_distribution)
+
+    kwargs = tf_supported_args[tf_distribution]
+    sample = kwargs.pop("sample", 0.1)
+
+    dist = _dist("test_dist", **kwargs, validate_args=True)
+
+    if tf_distribution is not "Binomial":
```
twiecki: str comparisons should use `!=`. Also, is this a TFP bug? If yes, should open an issue there and link to it here.

pymc4/tests/test_random_variables.py
twiecki: maybe rename to `test_rvs_logp_and_forward_sample`?

scripts/create_testenv.sh
```diff
@@ -0,0 +1,46 @@
...
+    fi
+fi
+
+
+# Install PyMC4 dependencies
+pip install --upgrade pip
+
+
+#  Install editable using the setup.py
+pip install  --no-cache-dir -r requirements.txt
```
twiecki:
```suggestion
pip install --no-cache-dir -r requirements.txt
```

scripts/create_testenv.sh
```diff
@@ -0,0 +1,46 @@
...
+pip install --no-cache-dir -r requirements-dev.txt
```
twiecki: missing new-line

pymc4/tests/test_random_variables.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+from .. import _random_variables
+
+
+@pytest.fixture(scope="session")
+def tf_supported_args():
+    """Provide arugments for each supported Tensorflow distribution"""
+    _tfp_supported_args = {
+        "Bernoulli": {"probs": 0.5},
```
twiecki: why not directly map the RV classes to their objects? Saves the `getattr` lookup below.

pymc4/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
+"""
+PyMC4 test configuration
+"""
+import pytest
+import tensorflow as tf
+
+
+@pytest.fixture(scope="function")
+def tf_session():
+    sess = tf.Session()
```
twiecki: How about:
```
with tf.Graph().as_default():
    with tf.Session() as sess:
        yield sess
```

pymc4/tests/test_random_variables.py
eigenfoo: There's no forward sampling in this function, right?

Also, no need to say `rvs`, it's in the random variable test module! Maybe `test_evaluating_log_prob`? 😄

canyon289: There is forward sampling I believe. It's why the binomial function is failing, I could use verification on this one thoug

pymc4/tests/test_random_variables.py
canyon289: I'll get rid of it

pymc4/tests/test_random_variables.py
canyon289: Will do both

scripts/Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM conda/miniconda3
+
+LABEL maintainer="Ravin Kumar"
```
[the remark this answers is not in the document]
canyon289: I just copied Austin from PyMC3 but happy to change it :)
https://github.com/pymc-devs/pymc3/blob/master/scripts/Dockerfile#L5

canyon289:
> Looks good!
> 
> One comment: there's a `binom_test.py` in the top-level directory: is that supposed to be there?

Whoops no. I'll delete it, thanks for catching that

pymc4/tests/test_random_variables.py
canyon289: This sounds way better. Will do

pymc4/tests/conftest.py
canyon289: Do we need to open and close session each time? I figured one session was fine if we cleared the graph, and opening and closing them would be a waste of computational resources, but I could be wrong

If we do implement this I'd prefer `sess.close()` in the finalizer since it'll make it easier to avoid stacked context managers, which then makes debugging easier and mocking easier imo

pymc4/tests/test_random_variables.py
canyon289: I remember why I didn't do this.

I wanted a couple things
1. An ordered object of distribution strings so I can pass them to the id argument function
2. A mapping of parameters to distributions
3. A guarantee that when testing a method, the id string matches the distribution being tested
4. Only one place where order is defined

If I mapped directly from rv.classes to their parameters the tests would still work, but I wouldn't have a way to get the right distribution name string to the id argument.

The other reason I felt this might be better is that when testing the distribution name is visible, even in pdb. If I mapped RV Classes, in a debugger they would show up as memory addresses since we haven't defined a `__repr__` for `RandomVariable` yet

There also is the possibility I'm wrong or explaining things poorly so let me know if this doesn't make sense.

pymc4/tests/test_random_variables.py
```diff
@@ -0,0 +1,97 @@
...
+        "Triangular": {"low": 0.0, "high": 1.0, "peak": 0.5},
+        "Uniform": {"low": 0, "high": 1},
+        "VonMises": {"loc": 0, "concentration": 1},
+        "Wishart": {"df": 3, "scale_tril": [[1]], "sample": [[1]]},
+    }
+
+    return _tfp_supported_args
+
+
+def test_normal_dist(tf_session):
```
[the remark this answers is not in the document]
canyon289: Yep

ferrine:
> > What conda channels do we plan on using? It appears the conda docker is only pulling from defaults, but we may want to grab things (tensorflow?) from conda-forge.
> 
> Fair point, right now we had some discussions about using `tf-nightly` and `tfp-nightly` and then switching to tf2 when tfp becomes compatible.
> 
> Besides those to requirements though nothing is installed by conda specifically, other than the base env packages. Everything else is done by pip and pypi

Should we then develop on top of eager mode?

pymc4/tests/conftest.py
ferrine: I'd vote for the current solution with graph reset. There is no difference between them. I see no computational resources waste...

canyon289:
@eigenfoo referenced this ticket somehow but I can't figure it out  so adding this reference as well

https://github.com/tensorflow/probability/issues/81

pymc4/tests/conftest.py
twiecki: @ferrine this suggests they are different: https://stackoverflow.com/questions/42706761/closing-session-in-tensorflow-doesnt-reset-graph

twiecki: I think it's quite easy to run into memory problems here if we keep adding to graph or create multiple sessions in these tests.

pymc4/tests/test_random_variables.py
twiecki: I don't follow that reasoning. 
> An ordered object of distribution strings so I can pass them to the id argument function
This isn't ordered once you created it as it's a dict. You could sort it but I don't see why you would want to do that.
> A mapping of parameters to distributions
This is a mapping of parameters to strings of distributions.
> A guarantee that when testing a method, the id string matches the distribution being tested
I don't see the benefit of that. You only increase the chance of having typos in the strings.
> Only one place where order is defined. If order is defined in two places, then coordination of order is human reliant, which means it'll definitely get messed up.
Where is order defined here in the first place? And why would that be important?

> If I mapped directly from rv.classes to their parameters the tests would still work, but I wouldn't have a way to get the right distribution name string to the id argument.
That's true, but do we need that?
> The other reason I felt this might be better is that when testing the distribution name is visible, even in pdb.
We do have the class name that will get printed.

In sum, I don't really see the benefit, other than easier lookup when trying to get parameters manually but I don't see the use-case as it's just a test. But I do think it's more brittle to have that level of indirection.

twiecki: OK, that looks way nicer. I wonder if we should just change the `__repr__` instead though.

canyon289: I think I reformulated this to meet the ask. Let me know what you think!

twiecki:
Thanks!

eigenfoo:
btw @canyon289 I just noticed this but fun fact, `sed` doesn't need to have a `/` follow the `s`: it can be any character so long as you're consistent. So 

`'s/\/opt\/pymc4/\/home\/travis\/build\/pymc-devs\/pymc4/g'`

could also just be

`'s_/opt/pymc4_/home/travis/build/pymc-devs/pymc4_g'`

which is more readable!

Issue #50, opened 2019-01-14
[Question] Where is object ArrayStep defined?


canyon289:
I can't seem to find the definition of `ArrayStep` in pymc4. Was this object experimental or is it supposed to be somewhere?

https://github.com/pymc-devs/pymc4/blob/master/pymc4/_hmc/hmc.py#L3

twiecki:
I think this is WIP from @aseyboldt. The idea is to copy our samplers, rip out the pymc3/theano stuff and make it work here. But as that part is not done there are still references from pymc3.

canyon289:
Thank you

PR #51, opened 2019-01-14
Just pylint

canyon289:
Splitting out pylint from testing. The issues with pytest might take a while to fix so might as well merge pylint in the meantime

### DIFF
--- a/pymc4/_hmc/hmc.py
+++ b/pymc4/_hmc/hmc.py
@@ -1,6 +1,7 @@
 import numpy as np
 
-from ..arraystep import Competence
+# Array step doesn't seem to exist
+# from ..arraystep import Competence
 from pymc3.vartypes import discrete_types
 from pymc3.step_methods.hmc.integration import IntegrationError
 from pymc3.step_methods.hmc.base_hmc import BaseHMC, HMCStepData, DivergenceInfo
--- a/pymc4/_hmc/hmc.py
+++ b/pymc4/_hmc/hmc.py
@@ -1,7 +1,5 @@
 import numpy as np
 
-# Array step doesn't seem to exist
-# from ..arraystep import Competence
 from pymc3.vartypes import discrete_types
 from pymc3.step_methods.hmc.integration import IntegrationError
 from pymc3.step_methods.hmc.base_hmc import BaseHMC, HMCStepData, DivergenceInfo
--- a/pymc4/_hmc/nuts.py
+++ b/pymc4/_hmc/nuts.py
@@ -5,7 +5,8 @@ from collections import namedtuple
 import numpy as np
 import numpy.random as nr
 
-from ..arraystep import Competence
+# Array step doesn't seem to exist
+# from ..arraystep import Competence
 from .base_hmc import BaseHMC, HMCStepData, DivergenceInfo
 from .integration import IntegrationError
 from pymc3.backends.report import SamplerWarning, WarningType
--- a/pymc4/_hmc/nuts.py
+++ b/pymc4/_hmc/nuts.py
@@ -5,8 +5,6 @@ from collections import namedtuple
 import numpy as np
 import numpy.random as nr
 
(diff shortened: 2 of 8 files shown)


pymc4/_random_variables.py
```diff
@@ -4,10 +4,11 @@
 Implements random variables not supported by tfp as distributions.
 """
 
+# pylint: disable=undefined-all-variable
```
eigenfoo: Why is pylint complaining that `contexts` is undefined? It looks like its defined in the `_template_contexts` module.

pymc4/_hmc/hmc.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 
-from ..arraystep import Competence
+# Array step doesn't seem to exist
```
eigenfoo:
```suggestion
# FIXME Array step doesn't seem to exist
```

pymc4/_hmc/nuts.py
```diff
@@ -5,7 +5,8 @@
 import numpy as np
 import numpy.random as nr
 
-from ..arraystep import Competence
+# Array step doesn't seem to exist
```
eigenfoo:
```suggestion
# FIXME Array step doesn't seem to exist
```

pymc4/_random_variables.py
canyon289: When you put a pylint ignore at the top it'll ignore any instance of that lint check in the entire file.

In this case context it's fine, it's all the distributions in the `__all__` array that it's complaining about. It can't "see" the definition in the code through static analysis, which makes sense because they're generated dynamically

pymc4/_hmc/nuts.py
canyon289: I removed these based on Thomas's reply to here
https://github.com/pymc-devs/pymc4/issues/50

pymc4/_random_variables.py
eigenfoo: Oh I see! I think it'd be best to describe that as a docstring that actually. Could you add that as a comment?

canyon289: Is it a comment? The format is strictly defined by pylint as well, we don't have any flexibility there

eigenfoo: I'm sorry, I was unclear. I was trying to say that we should have a comment like

```suggestion
# pylint cannot "see" the random variable definitions in the code through static
# analysis, since they are generated dynamically. We disable the specific linter
# warning here.
#
# pylint: disable=undefined-all-variable
```

I thought it would be good for future development, but perhaps it's an obvious detail? 😕

canyon289: So it's definitely not obvious if you're asking :)

But I do think adding a comment might be too much because we're documenting pylint's behavior, and I suppose this is were I'm drawing the line with "know your tools"

The only precendent I can give is examples from PyMC3 and ArviZ where we don't add comments. And my other argument is that files can contain many pylint disables, commenting each would be a bit much.

https://github.com/pymc-devs/pymc3/blob/master/pymc3/__init__.py#L1
https://github.com/pymc-devs/pymc3/blob/master/pymc3/tests/test_distributions.py#L1184
https://github.com/pymc-devs/pymc3/blob/master/pymc3/math.py#L3
https://github.com/pymc-devs/pymc3/blob/master/pymc3/math.py#L22

eigenfoo: That's a convincing case if I've heard one! In which case we can go ahead and merge.

requirements.txt
```diff
@@ -1,2 +1,4 @@
-tf-nightly-2.0-preview
+tf-nightly
```
junpenglao: I thought we are going to develop on tf-nightly-2.0-preview?

canyon289: We can't, tfp-nightly and tf2 don't work together at the moment

PR #54, opened 2019-01-19
TST: test all rvs, not just tfp supported distributions

### LINKED ISSUE #52: TST: tests for random variables
I was just thinking about the test suite here - I think it makes sense to dynamically generate tests for each random variable to:

1. Instantiate it
2. `.sample` from it
3. Evaluate its `.log_prob`
4. Convert it `as_tensor`

In other words, we should try to pull a similar trick to how we dynamically wrap the tfp distributions as random variables, for the tests. I think has several advantages, (e.g. the tests write themselves!), but I'm unsure if this is best practice with tests... Any thoughts?

cc @canyon289 depending on how far you want to take #49, I'd be happy to do a follow up PR for tests.

eigenfoo:
Closes #52 

Building off of #49, this PR adds tests for all current distributions, not just those supported by tfp.

### DIFF
--- a/pymc4/_random_variables.py
+++ b/pymc4/_random_variables.py
@@ -10,7 +10,6 @@ Implements random variables not supported by tfp as distributions.
 from . import _template_contexts as contexts
 
 import sys
-import numpy as np
 import tensorflow_probability as tfp
 from tensorflow_probability import distributions as tfd
 
--- a/pymc4/_random_variables.py
+++ b/pymc4/_random_variables.py
@@ -8,11 +8,28 @@ Implements random variables not supported by tfp as distributions.
 from . import _template_contexts as contexts
 
 import sys
+import numpy as np
+import tensorflow_probability as tfp
 from tensorflow_probability import distributions as tfd
 
 
-# Must match tfp.distributions names exactly.
-__all__ = [
+# Random variables tfp does not support as distributions. We implement these
+# random variables.
+tfp_unsupported = [
+    "Constant",
+    "DiscreteUniform",
+    "HalfStudentT",
+    "LogitNormal",
+    "Weibull",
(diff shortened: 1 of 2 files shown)


eigenfoo:
Well this caught a few bugs, mostly to do with using the wrong dtype (e.g. tf.float32 instead of tf.float64). I'll try and debug this sometime this weekend.

@canyon289 is it possible to get Coveralls to not comment like that? With PyMC3 it would should up in the same place that the Travis CI is.

twiecki:
Disabled coverall comments.

canyon289:
I think we need to split apart the tests for the tfp implemented distributions from the ones that pymc4 implements because for the pymc4 implemented distributions we do need to check for correctness

eigenfoo:
Currently the following code snippet fails with a `NotImplementedError` for the `inverse_log_det_jacobian`, in `bijector.py`. I propose we just comment out the `DiscreteUniform` distribution (and its tests) until tfp implements the relevant functions. I'm working on making sure the other distributions are suitably tested.

```
In [1]: import tensorflow_probability as tfp
   ...: from tensorflow_probability import distributions as tfd
   ...:
   ...: x = tfd.TransformedDistribution(distribution=tfd.Categorical(probs=[0.5, 0.5]),
   ...:                                 bijector=tfp.bijectors.AffineScalar(shift=1),
   ...:                                 name='DiscreteUniform')
   ...:
   ...: x.log_prob(2)
```

twiecki:
Looks great, thanks!

We should open an issue with TFP about the problems we found.

eigenfoo:
TFP issue filed here: https://github.com/tensorflow/probability/issues/286

twiecki:
Nice, looks like there is a simple solution available.

PR #59, opened 2019-01-20
DOC: numpy-style docstrings

eigenfoo:
Adds pydocstyle to our Travis builds, and fixes any docstrings to conform to numpy convention.

I've ignored error codes D100 through D107, which all raise errors if there are missing docstrings. In due course we will (hopefully!) get rid of those ignored errors.

Also, I made a `setup.cfg` with the configuration settings for `pydocstyle`. It seems that `black` does _not_ allow configs in a `setup.cfg` file, and instead looks for a `pyproject.toml` file. It seems unsatisfying to have so many config files littering our top-level directory.

### DIFF
--- a/pymc4/__init__.py
+++ b/pymc4/__init__.py
@@ -1,4 +1,4 @@
-"""PyMC4"""
+"""PyMC4."""
 from ._model import *
 from ._random_variables import *
 
--- a/pymc4/_hmc/quadpotential.py
+++ b/pymc4/_hmc/quadpotential.py
@@ -476,7 +476,7 @@ class QuadPotentialFull(QuadPotential):
 
 
 def add_ADATv(A, v, out, diag=None, beta=0.0, work=None):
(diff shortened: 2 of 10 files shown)


eigenfoo:
There is currently a bug in black which causes pydocstyle errors: https://github.com/ambv/black/issues/355

Black currently sometimes adds blank lines after function docstrings, which violates PEP257, and raises D202 when we run pydocstyle. So I've ignored D202, pending a bug fix in black.

canyon289:
> There is currently a bug in black which causes pydocstyle errors: [ambv/black#355](https://github.com/ambv/black/issues/355)
> 
> Black currently sometimes adds blank lines after function docstrings, which violates PEP257, and raises D202 when we run pydocstyle. So I've ignored D202, pending a bug fix in black.

Yes this happens. Git ignores are fine, or disable in pydocstyle

PR #62, opened 2019-01-21
Setup

aloctavodia:
I was unable to install PyMC4, this fix it.


ericmjl:
Thanks, @aloctavodia, for catching my error!

ColCarroll:
LGTM once checks pass (@ericmjl can revert if I missed something! :) )

Issue #58, opened 2019-01-20
Distribution parameters


ericmjl:
TFP Normal uses NumPy's `loc` and `scale`.

PyMC3 used `mu` and `sigma` (or alternatively `tau` or `sd`). 

What are our thoughts on parametrization of distributions? Follow TFP only, or provide the alternative parametrizations as well? 

Speaking for myself here, I believe that it's better for UX to keep PyMC3 parametrizations - it means fewer things to change API-wise, for an end-user.

eigenfoo:
Just to note: it looks like PyMC4's API will be quite different from PyMC3's in the first place. E.g. defining a model looks like it happens through a decorator instead of a context manager. So I don't think we should shy away from API changes if they bring benefits in the backend.

That being said though, I did like how PyMC3 had predictable Greek parameters - and TFP's parameter names are fairly unfriendly (e.g. the Beta distribution's `concentration0` and `concentration1`). I'm just concerned that renaming parameters will involve a lot of boilerplate code on our part. Basically I'd like to have Greek parameter names, but I think we should worry about that after we have working, well-tested random variables 😇

fonnesbeck:
I think we should err on the side of consistency with PyMC 3 over TFP, where we have a choice.

ericmjl:
Agreed - if that means creating a few extra lines of code in the codebase per probability distribution, such that migration from PyMC3 to PyMC4 is smoother, I think it's worthwhile for our users (of which we are subsets as well).

canyon289:
IMO the ideal scenario would be being able to copy over pymc3 models with minimal changes

ericmjl:
I like the "model as a function" paradigm, btw - it fits the "generative model" kind of way of thinking, IMO. :smile:

twiecki:
We settled this to use greek symbols like in PyMC3.

ericmjl:
*greek names. Greek unicode symbols would make things super challenging! :smile:

PR #63, opened 2019-01-24
ENH: add plots using ArviZ

eigenfoo:
https://github.com/pymc-devs/pymc3/pull/3338 lays out an easy plotting framework: depend on ArviZ for everything!

I've opted to retain the old PyMC3 names for the plots, even though they disagree with ArviZ's naming.

### DIFF
--- a/pymc4/__init__.py
+++ b/pymc4/__init__.py
@@ -1,6 +1,7 @@
 """PyMC4."""
 from ._model import *
 from ._random_variables import *
+from .plots import *
 
 import tensorflow as _tf
 from ._template_contexts import get_context as _get_context
--- a/pymc4/plots.py
+++ b/pymc4/plots.py
@@ -1,0 +1,20 @@
+"""
+PyMC4 plots.
+
+Plots are delegated to the ArviZ library, a general purpose library for
+exploratory analysis of Bayesian models. See https://arviz-devs.github.io/arviz/
+for details of plots.
+"""
+
+import arviz as az
+
+
+autocorrplot = az.plot_autocorr
+compareplot = az.plot_compare
+forestplot = az.plot_forest
+kdeplot = az.plot_kde
+plot_posterior = az.plot_posterior
+traceplot = az.plot_trace
+energyplot = az.plot_energy
+densityplot = az.plot_density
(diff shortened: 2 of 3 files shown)


ColCarroll:
Could be a top level `plot.py` with just 

```python
__all__ = az.plots.__all__
```

eigenfoo:
If I'm not mistaken, @canyon289 and @fonnesbeck were in favor of keeping the old API. I'm happy to go either way, but we should probably discuss that.

twiecki:
@eigenfoo That was about distributions but certainly the argument can be extended to other parts of the API. I don't think we should do it for plotting because the way pymc3 currently does it is more or less deprecated (only kept for backwards compat, otherwise we'd drop it). So I think this is an opportune time to move away from that here.

What do others think?

canyon289:
I'm in agreement with Thomas here.

There is value in keeping existing PyMC3 notebooks and code running as is, especially since its a library that's in maintenance mode with an active deprecation plan.

I actually don't think we should include plotting at all in PyMC4 and leave it totally to ArviZ to maintain integration. This way there is only one obvious way to do things for PyMC4 plotting. 

Otherwise we'll end up with the an api where `pm.traceplot` and `az.plot_trace` return the exact same result

twiecki:
Seems like we have a consensus, thanks everyone, especially @eigenfoo.

canyon289:
> Just referencing [#58 (comment)](https://github.com/pymc-devs/pymc4/issues/58#issuecomment-456075780) and [#58 (comment)](https://github.com/pymc-devs/pymc4/issues/58#issuecomment-456108539)

I perhaps should add one more comment for some more clarity. I think maintaining  consistency with models is generally a good idea. PyMC3, (like Stan, emcee, Pyro), is a PPL and the model is the core sell of the library. I personally like PyMC3's syntax, so admittedly this is personal preference.

In contrast plotting is not, I personally don't see a compelling reason why PyMC4 should have an alternative syntax from any of the libraries. Especially when putting my ArviZ hat on the goal is to unify plotting across Bayesian libraries.

This discussion is a good one and I'm glad we're having it. Thanks for making things concrete @eigenfoo

PR #68, opened 2019-02-04
MAINT: refactor random variables

eigenfoo:
~WIP.~

This PR refactors the random variables module.

- Those pesky underscores are removed
- `random_variables` is now its own directory with its own `__init__.py`, where the random variables are split up into `continuous`, `discrete` and `multivariate`, as we have in PyMC3. As we support more and more random variables (and as we add more and more documentation), this structure will make our lives easier in terms of maintenance.

Now, the random variables are exposed through `pm`, as well as through `pm.random_variables`. E.g., you can access `pm.Normal` and `pm.random_variables.Normal`. I believe this is the same as what we have with PyMC3.


canyon289:
Will review in next 48 hours!

twiecki:
Great stuff!

Issue #69, opened 2019-02-05
Extending TFP with alternative parametrizations


ericmjl:
@csuter this question references #58, which is something I care deeply about - having APIs that match what a package's user base might consider idiomatic. (We all probably know about the fiasco going from py2 to py3.)

I can see how the TFP distributions API matched and extended `numpy.random` better than TF did for numpy, which is admirable! I wonder if alternative parametrizations might be allowed at some point? 

E.g.: in PyMC3, we used parameter names that were familiar to a statistician, e.g. "mu" and "sigma" rather than "loc" and "scale" (this admittedly breaks the general principle I usually advocate for; in this case, though pm3's user base would be more familiar with mu and sigma over loc and scale). 

More e.g.: for other distributions, we allowed multiple parametrizations so that users could use the one they are most familiar with. I can foresee an objection - it's more code to maintain (but once implemented, can be hardened by tests) - however, I think that from a user adoption and migration perspective, this would be really helpful!

ericmjl:
@csuter this is great to know! 

@fonnesbeck @eigenfoo @twiecki, I think this provides much more clarity. If we want to have an API that matches PyMC3's, it's on us to ensure that the appropriate parameterizations are provided.

I think I can get to work starting with the Normal distribution :smile:.

PR #70, opened 2019-02-05
proposal PR to re-implement distributions API to match PyMC3

ericmjl:
At this point, I'm putting this PR out for discussion. I do not consider it finalized - there are only some docstrings and no tests. I will be shaping the PR based on what we talk about here.

Discussion points:

- If we want to maintain the PyMC3 API, it will be on us to provide alternate parameterizations (#69). Should we? This PR provides just a simple example in the form of the Normal distribution, but is extensible to other distributions, which I'm happy to work on.
- We will need to add tests for the alternate parametrizations, likely; what would be the best way? 
- Opinionated me suggests that we add warnings to end-users: "don't use loc + sigma together", "don't use mu + scale together", not because it's wrong per se, but because it is a hybrid of two paradigms of thinking about a distribution, which introduces friction when reading and reasoning about code. I have not implemented the warnings, but it'd be trivial with the stdlib. Should we do this?

I invited a bunch of people whom I think would have opinions to be reviewers for the PR; @aloctavodia I was thinking a lot about your book and how much of a hassle it'd be to update all of the examples to be used with PyMC4 if we didn't have these alternate parameterizations ported over. I'm also thinking of @AustinRochford's old blog posts, and my own Bayesian Analysis Recipes repository, as examples I'd love to see maintained as frictionlessly as possible.


fonnesbeck:
My preference is to have one parameterization, and not multiple (PyMC and TFP). I think it gets confusing otherwise (e.g. `Normal(loc, sigma)`). I don't think mapping will be onerous, because it is not likely to change, particularly if we are thoughtful about it from the beginnig.

eigenfoo:
I agree that PyMC4 should support friendlier argument names than TFP, but I'm not sure if I like the idea of _alternative_ parameter names. I.e. I think we should support either (`mu` and `sigma`) or (`loc` and `scale`), but _not both_ (I would prefer the Greek names for continuity with PyMC3).

As @ericmjl points out, alternative names would lead to "mixing" parameter names (imagine `mu` and `scale`), which is neither idiomatic nor Pythonic (there should be one and only one way to do it). I can't think of any function in the PyData ecosystem that allows you to pass parameters through one of two named arguments, and have the outcome be the same (usually that only happens when arguments are being deprecated).

The obvious exception would be when the parameters stand for mathematically different things: e.g. parametrizing a Normal through its variance (`sigma`), or its precision (`tau`).

ericmjl:
From @fonnesbeck: 

> I don't think mapping will be onerous, because it is not likely to change, particularly if we are thoughtful about it from the beginning.

Agreed! 

My lessons from pyjanitor mirror what Chris said: as long as the base is stable, the "friendly API" can do the tedious work up-front for the end-user. 

And from @eigenfoo: 

> I would prefer the Greek names for continuity with PyMC3.

And so would I too! 

I'm glad there's discussion here on this. Let me propose a change by pushing up a few more distributions parametrized the same way as in PyMC3, and see how many of the TFP distributions can be wrapped directly vs. needs a separate implementation.

twiecki:
I agree, I think this will make porting easier and provide nicer API for users.

twiecki:
Good points. After writing my comment I also had the thought that diverging from TFP might not be too wise if we want to allow some form of interaction between the two.

eigenfoo:
@twiecki what kind of interactions do you mean?

ericmjl:
@twiecki @junpenglao thanks for chiming in! 

Thomas, from your experience, which of the 80/20 use case does mixing in lower-level theano stuff fall in for you? Amongst PyMC3 users that I've interacted with, you've had the most experience going beyond the PyMC3-provided API.

Another related question for both of you - how often would you foresee a user needing to replace a PyMC4-parameterized distribution with a TFP-parameterized distribution?

At least from how I've interacted with PyMC3, I think I see it the way @fonnesbeck does - it's safe to provide a statistician-friendly parameterization, as the distributions base is foreseeably stable into the future. Also thinking about mixing/matching distributions, because PyMC4 distributions are building on top of TFP distributions, and because PyMC4 aims to be the high-level API for TFP model building, it's probably more advantageous for user adoption to keep the PyMC3 parameterizations. 

To summarize the main issues brought up so far, and my own thoughts on them:

- Will the TFP distributions need to be mixed in to PyMC4 code? From my experience, probably very little. For me & PyMC3, with the exception of the multivariate normal fix, I have had not any need to use NumPy's distributions. That said, I am probably too high-level of a user of PyMC3 :smile:.
- Will the TFP distributions foundation shift under us in the future? Likely not; the design of the TFP API mimics NumPy's API very closely. This means there will be a very low probability of needing "maintenance PRs" on the distributions library.
- Will retaining the PyMC3 distribution parameterizations help with migration? I think the answer is a clear yes. (`import pymc4 as pm` with minimal changes is extremely attractive for adoption.) There is a definite clear win for user-centered design. Whether this trumps over developer-centered design, however, is still debatable.
- Should we go for user-centered design or developer-centered design? This is our question to answer. My experience, both as a user of PyMC3 and a developer of `pyjanitor`, tells me that these might not be mutually exclusive, if we are thoughtful about things - and if we are conscientious about documenting these thoughtful thoughts (and not just leaving them as history in the issue tracker).

lucianopaz:
I agree that these are all very fair points but I agree more with @junpenglao, if there is an error downstream in TFP, the traceback with a different parametrization from pymc4 would very confusing for developers, and in my opinion even more so for users. I think that we could shift to use TFP's single parametrization (_one parametrization to rule them all and in darkness bind them..._)

fonnesbeck:
My preference for the PyMC3 parameterization is due to the fact that, as a high-level package for statistical modeling, its user-facing API is most important. So while Chris Suter's point about semantics being paramount is well-taken, that has to be balanced by the fact that our users are statisticians, analysts, and various other non-computer scientists. Their goal is in translating a mathematical model into computer code, so having conventions that are close to the math is going to be preferable. Using the location/scale convention would be somewhat easier from a developers perspective, but we aren't necessarily working to optimize for that. 

I've used the SciPy stats library for almost as long as its existed, and I still don't like that it uses non-statistical names for its arguments. It may be satisfying for its developers to have it that way, but it makes me less inclined to use it because I have to adapt my thinking every time I use it. There are going to be a number of places our conventions will diverge from TFP (e.g. 4-space vs 2-space indenting), and that's just an ongoing challenge of the project. PyMC4 is going to look different than TFP, and that's part of the point (otherwise our users would just write TFP models).

junpenglao:
Maybe we reintroduce @twiecki's idea here: https://github.com/pymc-devs/pymc3/pull/3308

kyleabeauchamp:
I'm torn.  I will say I don't like loc/scale because mathematical is the language of statistics, so IMHO there's no benefit in abstracting it out.  However, I also want to minimize the pain / maintenance burden, and sticking with the raw TF is great there.  

If we can come up with a "alternative mathy parameterizations layer" that requires very minimal maintenance, I would be fine with that.

ferrine:
All the comments on this are fair from by perspective. I think we are more user oriented and should go on with it. 

My only concern about pymc3 naming is debugging. Once a user sees tensorflow graph (it is weird) with different naming scheme for some nodes it will be impossible to figure out where does the error come from. From my tensorflow experience it was nearly impossible even without these problems. So this negative change is marginal.

But from user side we get fast adoption to the new paradigm. This itself might be painful but with tfp naming it will be worse.

ericmjl:
> Maybe we reintroduce @twiecki's idea here: [pymc-devs/pymc3#3308](https://github.com/pymc-devs/pymc3/pull/3308)

@junpenglao I choked when I clicked that link! 😆

ericmjl:
> if there is an error downstream in TFP, the traceback with a different parametrization from pymc4 would very confusing for developers

@lucianopaz I hear you! I think there's a few ways we can address this, even with the PyMC3 parametrization in place.

1. We can add tests (or even better, use Hypothesis to generate tests) that intentionally error out, to test the boundaries of PyMC4. I have been experimenting with Hypothesis on pyjanitor (and on other projects before), and it's been great for weeding out errors that I didn't catch initially.
2. In-line docstrings, can help. I'll push up a few examples of what I think would be useful docstrings. One example: explicitly stating that "mu" maps to "loc" and "sigma" maps to "scale". This ties into the point I made earlier ("being conscientious about documenting these thoughtful thoughts").
3. Developer documentation can also help. For example, TFP's naming conventions corresponds to this. We write in a so-called "API policy" (e.g. like CuPy's: essentially "mirror NumPy, and document when it doesn't"), and then stick to it, being clear PyMC4 can't be everything to everybody, because it'll then be nothing to nobody.

> Once a user sees tensorflow graph (it is weird) with different naming scheme for some nodes it will be impossible to figure out where does the error come from.

@ferrine that's a great point, especially when trying to get the user-facing API so-called "done well". I think the docs will have to reflect the differences appropriately.

> that has to be balanced by the fact that our users are statisticians, analysts, and various other non-computer scientists. Their goal is in translating a mathematical model into computer code, so having conventions that are close to the math is going to be preferable. Using the location/scale convention would be somewhat easier from a developers perspective, but we aren't necessarily working to optimize for that.

I think my thoughts are mirrored here. 

-----

As said above, though, I don't think user-centered design and developer-centered design are necessarily mutually exclusive, though it is a tad more up-front work (and hopefully not much in the way of maintenance work) to get set on the right path. I'm working on a few more examples on my laptop right now, so I hope to clarify some of the aforementioned ideas through them. Will keep you all posted.

eigenfoo:
@ericmjl I agree that our documentation needs to reflect any parameter name changes that we decide on, whatever they end up being. I also like the idea of developer documentation to describe the guts of the PyMC4-TFP mappings.

I also like the direction that this PR is going, by hand-implementing the classes, instead of dynamically creating the "TFP-supported" random variables, as we currently do. This makes the modules balloon in size (in terms of line count), but I think it is more maintainable in the long run: the docs are embedded in the code, the mappings are clear and easy to read (both in documentation and in code), and future developers won't get confused by the dynamic creation of PyMC4 random variables. The only downside is that we need to invest a lot of up-front effort to actually write `continuous.py` the first time around, which I think is a price worth paying.

Finally, as to using Hypothesis to test: I recall @ColCarroll thought it was good but not suitable for PyMC4, but don't remember the reason. We should discuss that.

pymc4/random_variables/continuous.py
```diff
@@ -14,6 +14,72 @@
 from .random_variable import RandomVariable
 
 
+class HalfNormal(RandomVariable):
+    r"""
```
canyon289: Loving this docstring

ericmjl: Merely a copy over from PyMC3! :smile:

ColCarroll:
In a vaccuum, with respect to naming, I would stick to the `tfp` conventions, because:

1. It errs on the side of writing/maintaining less code
2. It is another layer of indirection
3. The project is in development, and supporting parameter mappings might make dev work at least marginally harder

That said, this change already exists, improves user experience, and has a dev working on it enthusiastically, all of which are gifts, and I wouldn't be upset to merge it when ready.

junpenglao:
So we are committed to this change? I am still quite concern in particularly regarding potential confusing error message. 
I would be much more happy with it if it is possible to intercept the error and with a custom message that provide the correct parameter hint/mapping

ColCarroll:
Also @eigenfoo, regarding hypothesis, I think my objections when trying to port pymc3 were:

- tests took too long
- many failures weren't "important" (often because of `float32` or precision failures at certain scales)
- I'm not familiar enough with the library to make it a proper unittest (i.e., reliably repeatable locally and on CI)

I think all of these problems could be overcome, but they are what I ran into.

canyon289:
The only reason I would be upset to merge this PR is because we only finished one distribution. Regardless of which path we choose, I think we should commit to all or nothing

halfway change < choice1 <choice2

twiecki:
> Maybe we reintroduce @twiecki's idea here: [pymc-devs/pymc3#3308](https://github.com/pymc-devs/pymc3/pull/3308)

Well well well  ;)

ericmjl:
> I would be much more happy with it if it is possible to intercept the error and with a custom message that provide the correct parameter hint/mapping

@junpenglao that's a great point. Would you be open to providing an example that errors out in a confusing way? That'll help me grok a better idea of where the error interception should occur.

junpenglao:
@canyon289 seeing the PR title change I though we are going all the way. But yeah agree with you that we definitely should commit to one style. 

Alternatively, I think we can also prepare the PR and wait a bit before committing - at least that case the switching would not get into the way of development too much.

ericmjl:
> Alternatively, I think we can also prepare the PR and wait a bit before committing - at least that case the switching would not get into the way of development too much.

No problem with that! Happy to work it through and keep the PR open. I'll be sure to merge in changes from master frequently.

pymc4/random_variables/continuous.py
```diff
@@ -246,7 +246,7 @@ def _base_dist(self, sigma, *args, **kwargs):
 
 class HalfStudentT(RandomVariable):
     r"""
-    Half Student's T log-likelihood
+    Half Student's T distribution.
```
eigenfoo: I hate to nitpick but these are technically random variables, not distributions...

junpenglao: Also nit: mixing style of `r"""` and `R"""`

ericmjl: Fixed the nits!

pymc4/random_variables/continuous.py
```diff
@@ -14,15 +14,516 @@
...
+    ========  ==========================================
+
+    Parameters
+    ----------
+    sigma : float
+        Scale parameter :math:`sigma` (``sigma`` > 0) (only required if ``tau`` is not specified).
+
+    Examples
+    --------
+    .. code-block:: python
```
ferrine: pymc3 code spotted

ericmjl: Thanks @ferrine, I'll be sure to check at the end!

ericmjl:
Ok, looks like black standardizes on small `r"""`.

fonnesbeck:
@ColCarroll @junpenglao We are not committed to this change--its very much up for discussion still.

pymc4/random_variables/continuous.py
```diff
@@ -52,41 +1294,22 @@ def _base_dist(self, *args, **kwargs):
...
+# tfp_supported = ["InverseGaussian"]  # not present in PyMC3. We can choose how to parameterize
 
 # Programmatically wrap tfp.distribtions into pm.RandomVariables
-for dist_name in tfp_supported:
-    setattr(
-        sys.modules[__name__],
-        dist_name,
-        type(dist_name, (RandomVariable,), {"_base_dist": getattr(tfd, dist_name)}),
-    )
+# for dist_name in tfp_supported:
```
twiecki: These should just be deleted then rather than commented.

ericmjl: Thanks @twiecki! Resolved.

majidaldo:
Wouldn't it be possible to rename nodes on the graph?

twiecki:
I guess we should decide what we want to do here. Vote up/down on this item what you think.

twiecki:
@ericmjl I updated the comment to clarify.

pymc4/random_variables/continuous.py
```diff
@@ -67,7 +67,7 @@ class Beta(RandomVariable):
     - beta: concentration1
     """
 
-    def _base_dist(self, alpha, beta, *args, **kwargs):
+    def _base_dist(self, alpha: float, beta: float, *args, **kwargs):
```
ferrine: Why so constraining typehints here? Pycharm will always complain

ferrine: Why so constraining typehints here, pycharm will always complain

ericmjl: @ferrine we can expand it! :smile: What would be the appropriate types here? I'm thinking float, np.ndarray and tf.tensor?

ferrine: Do we even need them?

canyon289: It would help with code understanding and debugging as the library would get more complex

majidaldo: > It would help with code understanding and debugging as the library would get more complex

Perhaps. But because there are so many valid inputs (like float, [float], [[float1,] ,[float2]], np.array/tf.Tensor ), I'd suggest only the basic numeric types: float or int.

Keep in mind that there could be validation in the TFP Distribution.

ericmjl: Leaving this as a reminder for myself: I think I can do a thing where using the Typing module, we can create a new Type hint with an alias of some kind. 

```
Vector = List[float]
```

Could do something like:

```
TensorLike = Union[List[float], float, np.ndarray, tf.Tensor]
```

That would shorten the type hint and still be informative.

junpenglao: Alternatively we can force it to be tf.Tensor - as far as I understand, every will eventually convert to tf tensor in the execution at somepoint.

canyon289: > > It would help with code understanding and debugging as the library would get more complex
> 
> Perhaps. But because there are so many valid inputs (like float, [float], [[float1,] ,[float2]], np.array/tf.Tensor ), I'd suggest only the basic numeric types: float or int.
> 
> Keep in mind that there could be validation in the TFP Distribution.

The typing doesn't add anything for validation during runtime, it makes the code easier to read by humans and statically analyze by tools like mypy or pycharm

twiecki: If `tf.Tensor` will work just fine everywhere we should just use that. If not then I like @ericmjl proposal.

ericmjl: Because the type hints are for end-users and developers, and not for actual type checking, I think I will implement the TensorLike type hint.

pymc4/random_variables/continuous.py
```diff
@@ -519,7 +525,7 @@ def model():
     In PyMC3, HalfStudentT's location was always zero. However, in a future PR, this can be changed.
     """
 
-    def _base_dist(self, nu: int, sigma: float, *args, **kwargs):
+    def _base_dist(self, nu: IntTensorLike, sigma: TensorLike, *args, **kwargs):
```
ferrine: I remember pymc3 accepts float type as well

pymc4/random_variables/continuous.py
```diff
@@ -5,45 +5,1293 @@
...
 import tensorflow_probability as tfp
+import tensorflow as tf
 from tensorflow_probability import distributions as tfd
 
 from .random_variable import RandomVariable
+from typing import NewType, Union, Sequence
+import numpy as np
+
+
+TensorLike = NewType("TensorLike", Union[Sequence[int], Sequence[float], int, float])
```
ferrine: Why tf.Tensor is not included? Is it redundant?

ericmjl: Sequence[float] and Sequence[int] covers all Tensor containers, IIRC.

ericmjl:
Save for any final checks, I consider this PR done, as per discussion during today's dev meeting. cc: @twiecki.

Summary of changes:

- Ported pm3 docstrings.
- Ported pm3 arguments to match mathematical conventions.
- Type hints added.

pymc4/random_variables/types.py
```diff
@@ -0,0 +1,12 @@
+"""
```
twiecki: should this live in `random_variables`?

pymc4/random_variables/mixture.py
```diff
@@ -0,0 +1,77 @@
+from .random_variable import RandomVariable
+from tensorflow_probability import distributions as tfd
+import pymc4 as pm
+from .types import TensorLike
```
canyon289: Why does types need it's own module? In other words what does the layer of indirection add?

canyon289: scratch this, I see what you did.

ericmjl: It's so I don't have to re-implement the same types in each module, mostly.

pymc4/random_variables/types.py
ericmjl: That could work too instead.

pymc4/random_variables/mixture.py
ericmjl: But @twiecki's suggestion above is a good idea.

pymc4/random_variables/types.py
twiecki: Sorry, I actually like that it's a separate file, but I don't see why it should be in the subdir.

ericmjl:
Ok! All checks pass, and linear regression test is now done with the mathematically-oriented API.

ColCarroll:
Thanks a bunch, @ericmjl !
math

pymc4/tests/test_template_context.py
```diff
@@ -28,7 +28,7 @@ def test_free_forward_context_add_variable(monkeypatch):
     )
 
     with pytest.raises(Exception) as err:
-        random_variables.Normal("test_normal", loc=0, scale=1)
+        random_variables.Normal("test_normal", mu=0, sigma=1)
```
canyon289: I know I'm late to the merge by why sigma instead of sd like pymc3?

https://docs.pymc.io/api/distributions/continuous.html#pymc3.distributions.continuous.Normal

ericmjl: They were three equivalent parameterizations. I think I chose to put in sigma because it matches the greek mathematical notation. I was used to both before (for whatever reason, sometimes sd was more natural, and at other times, sigma was more natural).

twiecki: Yes, 100% sigma. I made that mistake originally in pymc3 and it's just not consistent. Either mean and sd, or mu and sigma, or loc and scale, but never a mix.

PR #71, opened 2019-02-06
TST: test backend arithmetic

eigenfoo:
Adds a short test to check that the arithmetic implemented by the `WithBackendArithmetic` class do not fail.

I removed the several arithmetic operations that failed because they aren't defined on `Tensor`s. E.g., you can't compute `divmod(x, y)` when `x` and `y` are Tensors.

Similarly for `x >> 1` and `x << 1`. I'm  not even sure if it makes sense for a Bayesian modeling library to support bitshifts.

### DIFF
--- a/pymc4/tests/test_random_variables.py
+++ b/pymc4/tests/test_random_variables.py
@@ -100,13 +100,13 @@ def test_rvs_backend_arithmetic():
     x = random_variables.Normal("x", 0, 1)
     y = random_variables.Normal("y", 1, 2)
 
-    assert x + y
-    assert x - y
-    assert x * y
+    assert x + y is not None
+    assert x - y is not None
+    assert x * y is not None
     # TODO test __matmul__ once random variables support shapes.
-    # assert x @ y
-    assert x / y
-    assert x // y
-    assert x % y
-    assert x ** y
-    assert -x
+    # assert x @ y is not None
+    assert x / y is not None
+    assert x // y is not None
+    assert x % y is not None
+    assert x ** y is not None
+    assert -x is not None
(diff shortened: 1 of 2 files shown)


pymc4/tests/test_random_variables.py
```diff
@@ -93,3 +93,20 @@ def test_rvs_logp_and_forward_sample(tf_session, randomvariable, kwargs):
         with pytest.raises(NotImplementedError) as err:
             dist.log_prob()
             assert "NotImplementedError: sample_n is not implemented: Binomial" == str(err)
+
+
+def test_rvs_backend_arithmetic():
```
canyon289: Not requiring a change but I think this would be more nicely tested in a parameterized test using lambda functions

Pros:
Each test would be split out by pytest and tested independently

Cons:
Its more code, less "direct", harder to read

eigenfoo: @canyon289 I thought about doing that, but the arithmetic is all implemented from the same backend class: I wasn't sure if it made sense to test the arithmetic for every random variable independently. I'm open to changing it, I'm just not sure what best test suite practices are.

canyon289: Ideally each test would test only one thing. In this case if the + operator fails it the test would stop and none of the other operators would be tested.

But it's really not too big of a deal. Feel free to merge if you feel like it :)

eigenfoo: Oh, I see! Should target for a follow up PR then, I suppose.

canyon289: I wouldn't worry about it. I think the time would be better spent elsewhere :)

PR #72, opened 2019-02-07
created makefile to for linting, testing, formatting

ericmjl:
cc @eigenfoo @canyon289 

This refactors the scripts into makefile commands. 

Need help getting the commands right for the docker containers in .travis.yml, though.


Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+
+style:
+	echo "Checking code style with pylint..."
+	python -m pylint pymc4/
+	echo "Success!"
+
+
+test:
+	pytest -v pymc4/tests/ --cov=pymc4/
```
canyon289: Would be nice to add html output coverage output here for non computer users

Makefile
```diff
@@ -0,0 +1,25 @@
+docstyle:
```
canyon289: Do a `phony` target need to be added to all of these?

https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Makefile
```diff
@@ -0,0 +1,25 @@
...
+	echo "Success!"
+
+
+test:
+	pytest -v pymc4/tests/ --cov=pymc4/
+
+lint: docstyle format style
+
+
+check: lint test
```
canyon289: Deleting `lint.sh` is going to be break CI. `.travisyml` file will need to be updated as well

ericmjl: Yep, exactly what I expected. I have identified the lines that I need to modify:

```
  - if [[ $NAME == UNIT ]]; then docker run  --mount type=bind,source="$(pwd)",target=/opt/pymc4 --name pymc4_test pymc4:latest bash -c \
                                              "pytest -v pymc4/tests/ --cov=pymc4/";
    fi

  - if [[ $NAME == LINT ]]; then docker run pymc4:latest bash -c "./scripts/lint.sh";
    fi
```

How would you modify these? I'm not quite well-versed with Docker.

canyon289: >   - if [[ $NAME == LINT ]]; then docker run pymc4:latest bash -c "make lint";
>     fi

That should work. You can leave the other one for now it'll still function

Makefile
ericmjl: I don't think so - regardless of whether files on disk have changed or not, make check works reliably.

Makefile
ericmjl: Do you have any opinions/prior knowledge on the best directory to store the output? 

We can generate a "testing-report.html", and add it to gitignore so it doesn't accidentally get checked into source control, perhaps.

canyon289: The output isn't just one html file, I think its a whole directory of html files. Adding it to `.gitignore`. is a great idea

Makefile
```diff
@@ -0,0 +1,25 @@
+docstyle:
+	echo "Checking documentation with pydocstyle..."
+	python -m pydocstyle ${SRC_DIR}/pymc4/
+	echo "Success!"
+
+
+format:
+	echo "Checking code style with black..."
+	python -m black -l 100 --check pymc4/
```
eigenfoo: Among the CI tools we use to check code, black is the only one that offers to in-place format your code for you. I'd like to take advantage of that: what do you think of having `make black` run `black -l 100 pymc4/` (without the `--check` flag)?

canyon289: We would need both a check and a format make command if we do this. That way the CI engine can check for black formatting without modifying the code

eigenfoo: Right, currently all recipes only check and don't modify code: I'm suggesting a new recipe (`make black` perhaps?) to do what I'm describing.

canyon289: We also could simplify this to

`black --check pymc4/`

Makefile
```diff
@@ -17,7 +17,7 @@ style:
 
 
 test:
-	pytest -v pymc4/tests/ --cov=pymc4/
+	pytest -v pymc4/tests/ --cov=pymc4/ --html=testing-report.html --self-contained-html
```
canyon289: This is cool!

Makefile
ericmjl: See commit: DA MASTAH BOMB FOMATTAH

Makefile
ericmjl: Solved.

Makefile
ericmjl: We're good - I figured out how to make it self-contained, with no assets directory necessary.

Makefile
ericmjl: Added a "black" command, and "check" now just runs black to format the code. "Check" is designed for a pre-commit check.

Makefile
ericmjl: Yay!

ericmjl:
@canyon289 @eigenfoo do let me know what else you think needs to be changed :smile:.

canyon289:
I'm just waiting for CI

eigenfoo:
Looks good, thanks @ericmjl!

Makefile
canyon289: Forgot to simplify this but nothing really changes ¯\_(ツ)_/¯

Was going to check this when I rereviewed but review got dismissed

PR #77, opened 2019-02-09
ENH: improve Makefile

eigenfoo:
A few improvements:

1. Added a help command. Running `make` now gives you:

```
Usage:

make conda      Set up a conda environment for development.
make venv       Set up a Python virtual environment for development.
make docker     Set up a Docker image for development.
make black      Format code in-place using black.
make test       Test code using pytest.
make lint       Lint code using pydocstyle, black and pylint.
make check      Both lint and test code. Runs `make lint` followed by `make test`.
```

2. Added rules to set up development environments: `make conda`, `make venv` and `make docker`.
3. Changed from `echo` to `printf`. [According to this SO post](https://stackoverflow.com/a/14121245), `make` is a bit treacherous with `echo`, and _may_ print newlines as `\n` depending on your shell or OS. Changing them all to `printf` is the simplest solution.
4. Removed `black` from the recipe for `make check`: `make check` should only check your code, not modify it!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,28 +12,28 @@ help:
 
 conda:  # Set up a conda environment for development.
 	@printf "Creating conda environment...\n"
-	${CONDA} create --yes --name env python=3.6
+	${CONDA} create --yes --name pymc4-env python=3.6
 	( \
-	${CONDA} activate env; \
+	${CONDA} activate pymc4-env; \
 	${PIP} install -U pip; \
 	${PIP} install -r requirements.txt; \
 	${PIP} install -r requirements-dev.txt; \
 	${CONDA} deactivate; \
 	)
-	@printf "\n\nConda environment created! \033[1;34mRun \`conda activate env\` to activate it.\033[0m\n\n\n"
+	@printf "\n\nConda environment created! \033[1;34mRun \`conda activate pymc4-env\` to activate it.\033[0m\n\n\n"
 
 venv:  # Set up a Python virtual environment for development.
 	@printf "Creating Python virtual environment...\n"
-	rm -rf venv
-	${PYTHON} -m venv venv
+	rm -rf pymc4-venv
+	${PYTHON} -m venv pymc4-venv
 	( \
-	source venv/bin/activate; \
+	source pymc4-venv/bin/activate; \
 	${PIP} install -U pip; \
 	${PIP} install -r requirements.txt; \
 	${PIP} install -r requirements-dev.txt; \
 	deactivate; \
 	)
-	@printf "\n\nVirtual environment created! \033[1;34mRun \`source venv/bin/activate\` to activate it.\033[0m\n\n\n"
+	@printf "\n\nVirtual environment created! \033[1;34mRun \`source pymc4-venv/bin/activate\` to activate it.\033[0m\n\n\n"
 
 docker:  # Set up a Docker image for development.
 	@printf "Creating Docker image...\n"
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ venv:  # Set up a Python virtual environment for development.
 docker:  # Set up a Docker image for development.
 	@printf "Creating Docker image...\n"
 	${SHELL} ./scripts/container.sh --build
-	@printf "\n\nDocker image created! \033[1;34mRun \`source venv/bin/activate\` to activate it.\033[0m\n\n\n"
+	@printf "\n\nDocker image created! \033[1;34mRefer to pymc4/scripts/README.md for instructions on how to use it.\033[0m\n\n\n"
 
 docstyle:
 	@printf "Checking documentation with pydocstyle...\n"
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ format:
 style:
 	@printf "Checking code style with pylint...\n"
 	pylint pymc4/
-	@printf "\033[1;34mPylint passes!\033[0m\n"
+	@printf "\033[1;34mPylint passes!\033[0m\n\n"
 
 black:  # Format code in-place using black.
 	black pymc4/
--- a/Makefile
+++ b/Makefile
@@ -62,4 +62,4 @@ test:  # Test code using pytest.
 
 lint: docstyle format style  # Lint code using pydocstyle, black and pylint.
 
-check: lint test  # Lint and test code.
+check: lint test  # Both lint and test code. Runs `make lint` followed by `make test`.
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,43 @@
+.PHONY: help venv conda docker docstyle format style black test lint check
+.DEFAULT_GOAL = help
+
+PYTHON = python3
+PIP = pip3
+CONDA = conda
+SHELL = bash
+
+help:
+	@printf "Usage:\n\n"
+	@grep -E '^[a-zA-Z_-]+:.*?# .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[1;34mmake %-10s\033[0m %s\n", $$1, $$2}'
+
+conda:  # Set up a conda environment for development.
+	@printf "Creating conda environment...\n"
+	${CONDA} create --name env
+	( \
+	source env/bin/activate; \
+	${CONDA} install -r requirements.txt; \
+	${CONDA} install -r requirements-dev.txt; \
+	deactivate; \
+	)
+	@printf "\n\nConda environment created! \033[1;34mRun \`source env/bin/activate\` to activate it.\033[0m\n\n\n"
+
+venv:  # Set up a Python virtual environment for development.
+	@printf "Creating Python virtual environment...\n"
+	${PYTHON} -m venv venv
+	( \
+	source venv/bin/activate; \
+	${PIP} install -U pip; \
+	${PIP} install -r requirements.txt; \
+	${PIP} install -r requirements-dev.txt; \
+	deactivate; \
+	)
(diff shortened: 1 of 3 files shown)


kyleabeauchamp:
Wow, what a great makefile!

eigenfoo:
Hm, it looks like `black`, `tf-nightly` and `tfp-nightly` are not on conda-forge. Should we just forget about the `make conda` rule?

kyleabeauchamp:
So with `conda env create -f env.yaml` you can specify both conda sources and pip sources.  Not sure what we can do about that if we're just using a requirements.txt, however.

eigenfoo:
Ah, good to know! I'll look into that later tonight. Thanks!

kyleabeauchamp:
Example: https://github.com/kyleabeauchamp/SolutionFFBench/blob/master/environment.yml#L22

canyon289:
I would have make run it in two commands. Create the conda env then run the
pip install after activating. The conda file method didn't work the last
time I checked but I could be wrong

ericmjl:
:heart: the `make conda` command. I have been using a conda environment for development, rather than a venv one.

Makefile
```diff
@@ -1,26 +1,66 @@
...
+PYTHON = python
+PIP = pip
+CONDA = conda
+SHELL = bash
+
+help:
+	@printf "Usage:\n"
+	@grep -E '^[a-zA-Z_-]+:.*?# .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[1;34mmake %-10s\033[0m%s\n", $$1, $$2}'
+
+conda:  # Set up a conda environment for development.
```
ericmjl: With a conda env, it may be better to provide an environment file. I'm kind of in two minds about this, though, which is why I have not PR'd my environment spec file to master. Providing an environment spec file is a form of duplication. There's always "minimize duplication" as a principle to adhere to, yet there's also a case for "some duplication can be very enabling". I don't know which is better here.

eigenfoo: That's a good point: I wonder if its even wise for us to accommodate two different kinds of development environments. If we just standardized on one of them, that might be easier in the long run...?

Also (without trying to start a flame war), if I have a conda environment, does it matter if I `conda install numpy` or if I `pip install numpy`? Is there a meaningful difference? Why is it better to have a separate `env.yml`?

ericmjl: With the latest release of conda 4.6, I think it now knows to check pip to satisfy dependencies. Prior to 4.6, I think conda would just clobber over pip, basically, if it didn't see a conda-installed package as a dependency of another package.

ericmjl: > Why is it better to have a separate `env.yml`? [also `environment.yml`]

The primary reason is that the `environment.yml` file is akin to a Dockerfile: it's a version-controlled specification for how to build a self-contained conda environment.

I also adopted a bunch of bash tooling surrounding the `environment.yml` file. For example, I use `conda-auto-env` to automatically source an environment if it sees an `environment.yml` file in a directory that I navigate to. If the environment doesn't exist, it'll auto-create the environment based on the spec file.

More of my own conda hacks are detailed [on my blog](http://ericmjl.com/blog/2018/12/25/conda-hacks-for-data-science-efficiency/) as well.

canyon289: I'm fairly strongly opinionated that we should not have both, and feel that we should use stick with a `requirements.txt`

The duplication doesn't make things easier and its simple enough to use pip and conda together

Makefile
```diff
@@ -1,26 +1,66 @@
+.PHONY: help venv conda docker docstyle format style black test lint check
```
canyon289: If we're going to use `.PHONY` here I think we should be consistent and use it everywhere

Makefile
```diff
@@ -1,26 +1,66 @@
...
+
+help:
+	@printf "Usage:\n"
+	@grep -E '^[a-zA-Z_-]+:.*?# .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[1;34mmake %-10s\033[0m%s\n", $$1, $$2}'
+
+conda:  # Set up a conda environment for development.
+	@printf "Creating conda environment...\n"
+	${CONDA} create --yes --name env python=3.6
+	( \
+	source activate env; \
```
canyon289: `source activate` is out of favor now. Should be `conda activate` like on line 21

Makefile
eigenfoo: Are you suggesting we keep the `make conda` rule, but have it install from `requirements.txt` (no `environment.yml`), or are you suggesting we ditch the `make conda` rule altogether?

Makefile
```diff
@@ -1,26 +1,66 @@
...
 
 style:
-	@echo "Checking code style with pylint..."
+	@printf "Checking code style with pylint...\n"
 	pylint pymc4/
-	@echo "Pylint passes!\n"
+	@printf "\033[1;34mPylint passes!\033[0m\n\n"
 
-black:
+black:  # Format code in-place using black.
```
canyon289: I'm being so super nitpicky here but if we're commenting some of the make recipes we should comment them all. Not a big deal if we don't though since these recipes are really obvious

scripts/README.md
```diff
@@ -1,12 +1,12 @@
...
 
-* Image can be built locally using command
+* Image can be built locally using the command `make docker` or the command
  `./scripts/container.sh --build` from the root `pymc4` directory
  
 * After image is built an interactive bash session can be run 
 `docker run -it pymc4 bash`
 
-* Command can be issused to the container such as linting and testing
+* Command can be issued to the container such as linting and testing
```
canyon289: Thanks for catching this!

canyon289:
This is great. Thanks @eigenfoo

Makefile
eigenfoo: That's what I thought as well! I decided not to document all of them though: if you document it, it will show up when you run `make help`, or just `make`. I only documented the rules that I expected end-users to actually use (in other words, I think developers should focus on "linting", instead of "linting with black" or "linting with pylint"). What do you think?

scripts/README.md
eigenfoo: :+1:

Makefile
eigenfoo: Gotcha!

Makefile
canyon289: My preference would be to have a `requirements.txt` or an `environment.yml` but not both. I think a `requirements.txt` is preferable because it doesn't require conda. 

Thats my 2 cents :)

Makefile
canyon289: Your explanation is sufficiently convincing for how minor my concern is

I think whatever you do here will be great :)

Makefile
eigenfoo: What do you mean by everywhere? In other Makefiles?

Makefile
eigenfoo: Agreed! I'll keep the `make conda` rule: it will just install via the `pip` in the conda environment, instead of using `conda install`.

Makefile
canyon289: On every target. For example line 9, line 13, line 38 etc.

Eric pointed out in his original PR that it wasn't needed, but I'm somewhat a stickler for consistency

canyon289: Or maybe, I just realized, I don't understand makefiles well enough and by specifying it on line one, that's the equivalent of specifying it for everything.........

eigenfoo: Ah I see! In Makefiles `.PHONY` is a special reserved variable that indicates what recipes are named (so that Make doesn't confuse them for file names), so specifying them up front is all you need to do (it doesn't even need to be on line 1).

Makefile
```diff
@@ -1,26 +1,66 @@
...
+CONDA = conda
+SHELL = bash
+
+help:
+	@printf "Usage:\n"
+	@grep -E '^[a-zA-Z_-]+:.*?# .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[1;34mmake %-10s\033[0m%s\n", $$1, $$2}'
+
+conda:  # Set up a conda environment for development.
+	@printf "Creating conda environment...\n"
+	${CONDA} create --yes --name env python=3.6
```
twiecki: shouldn't we call the environment `pymc4` instead?

Makefile
```diff
@@ -1,26 +1,66 @@
...
+	${PIP} install -r requirements.txt; \
+	${PIP} install -r requirements-dev.txt; \
+	${CONDA} deactivate; \
+	)
+	@printf "\n\nConda environment created! \033[1;34mRun \`conda activate env\` to activate it.\033[0m\n\n\n"
+
+venv:  # Set up a Python virtual environment for development.
+	@printf "Creating Python virtual environment...\n"
+	rm -rf venv
+	${PYTHON} -m venv venv
```
twiecki: Same -> `pymc4`.

eigenfoo: Python venvs make another directory to store the dependencies. It can't be called `pymc4` because there's already a `pymc4`! I could call it `pymc4-venv`.

twiecki:
Best makefile ever.

fonnesbeck:
I did notice that failed docker builds still return "Docker image created!" messages.

eigenfoo:
Docker is strange in that it always seems to exit with exit code 0... it's not easy for me to prevent that from happening. If anything I can just remove the "Docker image created!" message altogether.

canyon289:
> Docker is strange in that it always seems to exit with exit code 0... it's not easy for me to prevent that from happening. If anything I can just remove the "Docker image created!" message altogether.

I think removing the line is a good idea. Docker reports it's own errors, but if you want to "hoist" the exit code here's a helpful snippet

https://nickjanetakis.com/blog/docker-tip-22-checking-the-exit-code-of-stopped-containers

eigenfoo:
Taking the path of least resistance in #80 😄

PR #82, opened 2019-02-13
make DiscreteUniform work by typing with float32 #81

### LINKED ISSUE #81: ENH: add DiscreteUniform
#52 left out the `DiscreteUniform` distribution (it is commented out right now). I misunderstood the errors that TFP was raising, which was ironed out in https://github.com/tensorflow/probability/issues/286. All that remains is to properly implement the DiscreteUniform distribution (most of it is written, we just need to iron out the type casting).

majidaldo:
Closes #81

### DIFF
--- a/pymc4/random_variables/discrete.py
+++ b/pymc4/random_variables/discrete.py
@@ -9,6 +9,9 @@ Implements random variables not supported by tfp as distributions.
 # pylint: disable=undefined-all-variable
 import sys
 from tensorflow_probability import distributions as tfd
+import tensorflow_probability as tfp
+import tensorflow as tf
+import numpy as np
 
 from .random_variable import RandomVariable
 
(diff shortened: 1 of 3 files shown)


pymc4/random_variables/discrete.py
```diff
@@ -17,7 +20,7 @@ class Constant(RandomVariable):
     _base_dist = tfd.Deterministic
 
 
-''' TODO DiscreteUniform raises a NotImplementedError from tfp.
+# TODO DiscreteUniform raises a NotImplementedError from tfp.
```
canyon289: This comment should be removed now right?

majidaldo: ah ya. duh.

ericmjl:
It's also implemented in #70, but no worries - let's get this merged!

PR #85, opened 2019-02-20
WiP multivar skew normal

majidaldo:
#44 

Sorry about this being very rough but I found it challenging just to get to this point. I've only implemented sampling so far and, visually, it looks correct.

corr =  
[[1, .7],
 [.7, 1]]  ) # mat
skew =
[[.5],
[-5]]
image

I fried my brain trying to do this with a bijector but, in the end, it seemed I had to make a TFP distribution.

todo:
- [x] log_p
- [x] batching/broadcasting
- [x] use a bijector to locate and scale.
- [x] wrap with (PyMC) RV
- [ ] test


twiecki:
I wonder if this shouldn't go into TF probability instead?

pymc4/random_variables/continuous.py
```diff
@@ -42,6 +49,165 @@ def _base_dist(self, *args, **kwargs):
...
+    
+    def _event_shape_tensor(self):
+        return constant_op.constant([array_ops.shape(self.skew)[-1]], dtype=dtypes.int32)
+
+    def _event_shape(self):
+        return tensor_shape.TensorShape([self.skew.get_shape()[-1].value])
+    
+    def _sample_n(self, n, seed=None):
+        nd = self.event_shape[0] + 1
+        N = tfd.MultivariateNormalFullCovariance(loc=zeros(nd), covariance_matrix=self.get_Omega_star())
```
brianwa84: for structured covariance matrices like the one you're using here it might be much more efficient (i.e. save the N^3 cholesky operation that MVNFull does by default) if you can use `tf.linalg.LinearOperator` and `tfd.MVNLinearOperator`

pymc4/random_variables/continuous.py
```diff
@@ -42,6 +49,165 @@ def _base_dist(self, *args, **kwargs):
...
+    def get_Omega_star(self):
+        """correlation matrix of underlying normal distribution"""
+        delta = self.get_delta()
+        Omega = self.Omega
+        one_shp = array_ops.shape(delta)
+        # one_shp[-2] = 1 doesn't work
+        # https://github.com/tensorflow/tensorflow/issues/14132
+        # so instead, ...
+        one_shp = concat([one_shp[:-2] , [1,1] ], 0)
+        one = ones(one_shp)
```
brianwa84: you could do array_ops.shape(delta[..., :1, :1])

also generally outside TF we recommend using API references like tf.shape(..) instead of array_ops.shape(..)

majidaldo:
@brianwa84 isn't there a good amount of logic in the code that i submitted that should be in tfp? i feel like, as a distribution implementer, i shouldn't have to deal so much with broadcasting .

brianwa84:
Ideally you should not need to do explicit broadcasting in most cases -- TF ops should take care of it for you.

Perusing the code, I think it could benefit from renaming some things to avoid greek (you can still mention it if helpful in a comment `# maps to omega from eq 1.x in the paper`) and instead express what the variable is doing (skew matrix?)

We do want to eventually get a skew MVN in TFP, so you're welcome to move this PR over to our repo.

twiecki:
@majidaldo Closing this but encouraging you to open it on TFP.

PR #83, opened 2019-02-18
initial implementation of transforms

ericmjl:
Resolves #78.

Issue #79, opened 2019-02-11
Do not require passing of RV name by AST transformation


twiecki:
PyMC3 as well as Edward and Edward2 require (or strongly encourage) passing of an additional name to each RV:
```python
with pm.Model():
     x = pm.Normal('x', mu=0, sigma=1)
```
Note that the `x` appears twice. We need to do this because we can't easily infer that the variable `x` is called `"x"` at run-time. But we do need to keep track of it and reference it later in the trace, so that's why we need the additional argument.

This is in contrast to DSL PPL systems like stan, bugs or newer julia-based PPLs which don't need that.

It's also a common stumbling block and dislike for many pymc3 users:
![image](https://user-images.githubusercontent.com/674200/52556560-38812e80-2ded-11e9-94f9-cb7ba0b28d51.png)

![image](https://user-images.githubusercontent.com/674200/52556574-420a9680-2ded-11e9-8ec3-6a470022d360.png)

I hacked together a small prototype that inspects and transforms the AST to get the information and make the name *optional* (i.e. it can still be provided by the user):
![image](https://user-images.githubusercontent.com/674200/52557484-dbd34300-2def-11e9-9e75-98def949733f.png)

This is fairly straight-forward as you can see, we just fetch the name of the variable and add it to the code that instantiates the RV. So for pymc4 everything looks identical to before, it's just a thin layer that transforms the code before it get's executed. 

This prototype matches the PyMC3 API but I think we should change it ultimately to look something like this:
```python
def create_RV():
    return pm4.Normal(mu=0, sigma=1, name='required') # name must be provided as we don't traverse into function calls, although we probably could

@pm4.model(autoname=True) # True by default?
def mymodel():
    x = pm4.Normal(mu=0, sigma=1) # replaced by AST parser to -> x = pm4.Normal(mu=0, sigma=1, name='x')
    x2 = pm4.Normal(mu=0, sigma=1, name='but_i_want_to') # name can be supplied
    y = create_RV() 
    z = [pm4.Normal(mu=0, sigma=1, name='required {}'.format(i) for i in range(5)] # name must be provided
    x = pm4.Normal(mu=0, sigma=1) # raises exception, variable already defined
```

That code focuses mostly on edge cases where we can't easily infer the name, we don't need to handle those cases (as I think they are rare).

[...]

PR #86, opened 2019-03-02
Auto name RVs by parsing AST

twiecki:
Based on #83. See #79 for more discussion.

image

This code directly rewrite the bytecode of the function which is quite elegant because it all looks and runs the same. For example, debugging still works:
image

We can also easily detect cases where we can't infer the name and raise an exception that no name was provided.

### DIFF
--- a/pymc4/tests/test_autoname.py
+++ b/pymc4/tests/test_autoname.py
@@ -2,21 +2,26 @@
 Tests for PyMC4 auto naming.
 """
 
+# AST parsing only works in non-nested functions, thus place things into module namespace
 import pymc4 as pm
+
+# AST looks different if we use pm.Normal or just Normal
 from pymc4 import Normal
 
 
 def create_rvs(**kwargs):
-    return pm.Normal(0, 1, name="test")
+    # AST parsers should just add name kwargs to pymc4 RVs
+    assert "name" not in kwargs.keys()
+    return pm.Normal(0, 1, name="inside_function")
 
 
 @pm.model(auto_name=True)
 def autoname_test():
-    mu = pm.Normal(0, 1)
-    mu2 = Normal(0, 1)
-    mu3 = pm.random_variables.Normal(0, 1)
-    pm.Normal(0, 1, name="custom")
-    dummy = pm.Normal(0, 1, name="custom2")
+    inferred_w_module = pm.Normal(0, 1)
+    inferred_wo_module = Normal(0, 1)
+    inferred_full_path = pm.random_variables.Normal(0, 1)
+    pm.Normal(0, 1, name="supplied_name")
+    dummy = pm.Normal(0, 1, name="overwrite_name")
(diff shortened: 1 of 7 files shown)


twiecki:
OK I think this is ready for review.

pymc4/tests/test_autoname.py
```diff
@@ -0,0 +1,30 @@
+"""
+Tests for PyMC4 auto naming.
+"""
+
+from .. import model
+
+import pymc4 as pm
```
canyon289: Can these be changed to relative imports for consistency?

pymc4/tests/test_autoname.py
```diff
@@ -0,0 +1,30 @@
...
+from pymc4 import Normal
+
+def create_rvs(**kwargs):
+    return pm.Normal(0, 1, name="test")
+
+
+@pm.model(auto_name=True)
```
canyon289: This could be placed in fixture to to keep global scope clean

pymc4/tests/test_autoname.py
```diff
@@ -0,0 +1,30 @@
...
+def autoname_test():
+    mu = pm.Normal(0, 1)
+    mu2 = Normal(0, 1)
+    mu3 = pm.random_variables.Normal(0, 1)
+    pm.Normal(0, 1, name="custom")
+    dummy = pm.Normal(0, 1, name="custom2")
+    x = create_rvs()
+
+
+def test_auto_name():
```
canyon289: This test really shows off the how nice the API is!

pymc4/tests/test_autoname.py
twiecki: Actually that doesn't work here as we can't parse AST nested functions. I added comments to explain that.

ericmjl:
@twiecki ready to merge?

twiecki:
yes!

twiecki:
Ugh, this method doesn't seem to work with autograph which is run with `defun` :(.

fonnesbeck:
Do you have an example of where this fails?

twiecki:
@fonnesbeck e.g.
 
```python
@pm.model(auto_name=True)
def model():
    x = [pm.Normal(0, 1) for i in range(5)]
    pm.Normal(0, 1)
```

twiecki:
> Ugh, this method doesn't seem to work with autograph which is run with defun :(.

I have a fix over here: https://github.com/pymc-devs/pymc4/pull/90/commits/3d55143c0da38e179075972f45ca025782bb717b it's not super clean but it does the trick.

brandonwillard:
Is the `defun` issue solved by changing the decorator order?

twiecki:
I don't think we can, as then we'd have to parse the super-mangled AST that `defun` leaves us. The current "hack" works quite well and I don't see a problem with it.

PR #90, opened 2019-03-10
WIP Transform RVs, not just their logp.

twiecki:
```python
@pm.model
def t_test():
    sigma = pm.HalfNormal('sigma', 1)
    mu = pm.Normal('mu', 0, sigma)

model = t_test.configure()
func = model.make_log_prob_function()
sd = -tf.ones((1,))
mu = tf.ones((1,))
func(sd, mu)
```

```
<tf.Tensor: id=325, shape=(), dtype=float32, numpy=-4.5149107>
```

### DIFF
--- a/pymc4/random_variables/random_variable.py
+++ b/pymc4/random_variables/random_variable.py
@@ -1,4 +1,4 @@
- """
+"""
 PyMC4 base random variable class.
 
 Implements the RandomVariable base class and the necessary BackendArithmetic.
@@ -10,7 +10,7 @@ Also stores the type hints used in child classes.
 
 from .. import _template_contexts as contexts
 from tensorflow_probability import distributions as tfd
-from tensorflow_probability import bijectors  # import Bijector
+from tensorflow_probability import bijectors
 from typing import NewType, Union, Sequence
 
 
@@ -114,7 +114,7 @@ class RandomVariable(WithBackendArithmetic):
 
     def __init__(self, *args, **kwargs):
         self._parents = []
-        self._untransformed_distribution = self._base_dist(name=name, *args, **kwargs)
+        self._untransformed_distribution = self._base_dist(*args, **kwargs)
         self._sample_shape = ()
         self._dim_names = ()
         ctx = contexts.get_context()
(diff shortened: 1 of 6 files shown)


pymc4/random_variables/random_variable.py
```diff
@@ -157,6 +159,15 @@ def log_prob(self):
         """
         return self._transformed_distribution.log_prob(self)
 
+    def as_tensor(self):
```
ericmjl: Can we leave a brief docstring here explaining why `as_tensor` is reimplemented here?

twiecki:
Added a new commit with a new idea: We *always* transform distributions, just in some cases that transformation is the identity. This will save us a whole bunch of extra checks and unify the whole thing.

pymc4/random_variables/random_variable.py
```diff
@@ -108,30 +108,39 @@ class RandomVariable(WithBackendArithmetic):
...
-        self._distribution = self._base_dist(name=name, *args, **kwargs)
+        self._untransformed_distribution = self._base_dist(name=name,
+                                                           *args, **kwargs)
         self._sample_shape = ()
         self._dim_names = ()
         self.name = name
         ctx = contexts.get_context()
         self._creation_context_id = id(ctx)
         self._backend_tensor = None
+        if "bijector" in kwargs.keys():
```
canyon289: Being picky here but you could change this to self._bijector = kwargs.get("bijector",  _bijector)

twiecki: we could do `self._bijector = kwargs.get("bijector", self._bijector)` if that's what you meant.

canyon289: Yea that :)

pymc4/random_variables/random_variable.py
```diff
@@ -140,62 +148,21 @@ def as_tensor(self):
         if self._backend_tensor is None:
             self._backend_tensor = ctx.var_as_backend_tensor(self)
 
-        return self._backend_tensor
+        return self._bijector.forward(self._backend_tensor)
 
 
 class ContinuousRV(RandomVariable):
```
canyon289: If this and `DiscreteRV` both do nothing let's get rid of them. The extra indirection doesn't help any

twiecki: Good point.

pymc4/random_variables/random_variable.py
ericmjl: nvm, I'm being dumb here.

ericmjl:
@twiecki I pushed up a few more changes. Please check them out.

twiecki:
@ericmjl Thanks! I accidentally removed them during a rebase but reimplemented them.

twiecki:
Superseded by https://github.com/pymc-devs/pymc4/pull/92.

PR #92, opened 2019-03-12
Fix transforms, fix NBs, tests, move us to TF2

twiecki:
This should get us back to a more or less stable state where NBs (mostly) run and tests (mostly) pass. 

I'm also running things on TF2 nightly now. You can install it via `pip install -q --upgrade tf-nightly-2.0-preview tfp-nightly`

This also fixes the issue with the transformed variables. 

I also renamed the NBs to give them more descriptive names.

I could need help with fixing up the tests that are still breaking (looks like casting issues).


pymc4/_hmc/integration.py
```diff
@@ -91,7 +84,7 @@ def _step(self, epsilon, state, out=None):
         # q_new = q + epsilon * v_new
         axpy(v_new, q_new, a=epsilon)
 
-        logp = self._logp_dlogp_func(q_new, q_new_grad)
```
twiecki: Is this correct? CC @aseyboldt

pymc4/_hmc/quadpotential.py
```diff
@@ -163,9 +160,9 @@ def __init__(
         self.dtype = dtype
         self._n = n
         self._var = np.array(initial_diag, dtype=self.dtype, copy=True)
-        self._var_theano = theano.shared(self._var)
+        #self._var_theano = theano.shared(self._var)
```
twiecki: Not sure where and how that is being used.

canyon289:
This is quite the PR. Will be going through this in detail over the next 3 days

canyon289:
@twiecki Can you post the script you used to get the posterior dist? Would like to dynamically inspect the code

twiecki:
it's the HMC ipynb

junpenglao:
It would probably easier to review if you can split the PR into two? Especially the port of PyMC3's HMC and NUTS sampler part.

twiecki:
OK, I spit out the HMC part to https://github.com/pymc-devs/pymc4/pull/98.

pymc4/_template_contexts.py
```diff
@@ -58,10 +60,13 @@ class InferenceContext(BaseContext):
 
     def __init__(self, tensors, expected_vars):
         self.vars = []
+        self._names = [var.name for var in expected_vars]
         self._tensors = {var.name: tensor for var, tensor in zip(expected_vars, tensors)}
```
canyon289: @twiecki It seems like we're going to have to be extra careful about passing in tensors and expected_vars in the correct order from wherever this is going to be called.

Is this a good design pattern? I honestly don't know, am just asking

twiecki: Yeah, the order has to keep constant in-between function calls. It's not ideal but also not totally crazy. I view this as a temporary work-around (i.e. hack) to get things working with `function()` which messes with the AST in its own way. There are different, more complex solutions to this, but they'd be harder to implement.

twiecki:
@canyon289 Thanks so much for fixing this up!!

pymc4/_template_contexts.py
```diff
@@ -58,10 +60,13 @@ class InferenceContext(BaseContext):
...
+        return
 
     def add_variable(self, rv):
         self.vars.append(rv)
+        return
```
twiecki: Is this black?

twiecki:
I'm merging this to get our master back in shape.

pymc4/_template_contexts.py
canyon289: No its me adding it so I have a place to activate the inline debugger after all the attribute have been instantiated.

pymc4/_template_contexts.py
canyon289: Just briefly what are those solutions?

twiecki: The problem is that our current method changes the byte code of the function, not the source. That's totally fine if you just execute it, but if you then do AST parsing on that again, like auto-graph does, it will still see the old source code (because we didn't touch that). Auto-graph does actually change the source code of the function (so that's possible) but it's more complex and requires writing a file out to disk. The machinery they're building could be used for our purposes as well, but this short-cut is just way simpler.

PR #98, opened 2019-04-09
Port HMC from pymc3

twiecki:
I ported PyMC3's samplers to here. There were quite a few theano and pymc3 references I pulled out, but then it was fairly straight forward. The HMC NB shows a very simple usage. I don't think it's working quite yet, the step_size gets smaller and smaller and more and more samplers are getting drawn, not sure if I messed the sampler up somehow. Here is a demo: https://github.com/pymc-devs/pymc4/blob/f9e83272d5accab792b82218e3feccf2d7a240c0/hmc.ipynb


pymc4/_hmc/integration.py
```diff
@@ -91,7 +84,7 @@ def _step(self, epsilon, state, out=None):
         # q_new = q + epsilon * v_new
         axpy(v_new, q_new, a=epsilon)
 
-        logp = self._logp_dlogp_func(q_new, q_new_grad)
+        logp, q_new_grad = self._logp_dlogp_func(q_new)
```
twiecki: I hope this is doing the right thing.

pymc4/_hmc/quadpotential.py
```diff
@@ -163,9 +160,9 @@ def __init__(
         self.dtype = dtype
         self._n = n
         self._var = np.array(initial_diag, dtype=self.dtype, copy=True)
-        self._var_theano = theano.shared(self._var)
+        #self._var_theano = theano.shared(self._var)
```
twiecki: Not sure what this was for.

twiecki:
We should probably also port PyMC3's tests for HMC and NUTS.

pymc4/_hmc/integration.py
ColCarroll: i'll have to poke around this later, but it looks like we're providing the arrays to write to, rather than making numpy/theano allocate them. Which is to say, if `_logp_dlogp_func` does what it looks like it does, this line should be 

```
logp, q_new_grad[:] = self._logp_dlogp_func(q_new)
```

the pymc3 implementation could more explicitly be `self._logp_dlogp_func(q_new, out=q_new_grad)`

ColCarroll:
thanks for starting this. will be interesting if we can get benchmarks on speed here vs. tfp's HMC.

twiecki:
@ColCarroll I fixed your suggestion. The problem with the decreasing step-size remains. I added that output to the NB: https://github.com/pymc-devs/pymc4/blob/1ec5de567e433b92163d4495670f99b248e68c61/pymc3_samplers.ipynb

ColCarroll:
this was a funny bug: we are using the BLAS function `axpy`, which computes `x = a * x + y` in place, and very fast, to do the leapfrog integration. It is a (good) micro-optimization. Something was going wrong with that here. It is probably worth a bug report to _someone_ if we can find a reproducible example.

I also got rid of the in place updating for now, since it is not being used anywhere in PyMC3 yet, anyways.

I think this branch is now ready to go -- if you check out the (very helpful) notebook you made, the two implementations are nearly identical now.

image

ColCarroll:
I've also got a few refactors I'd like to make after looking so carefully at the step size adaptation, but will wait until we merge this!

twiecki:
@ColCarroll Amazing, thanks so much for your help!

twiecki:
I just tried the NUTS port and it works as well. This is amazing progress.

PR #99, opened 2019-04-09
Fix HMC branch

kyleabeauchamp:
1.  This fixes the linter warning/error on travis that says "--py36 is deprecated and will be removed in a future version. Use --target-version py36 instead."
2.  Also fix use of un-imported `DiscreteRV` that was causing UndefinedVariable exception.

PR #100, opened 2019-04-10
Fix black settings in pyproject

kyleabeauchamp:
As was done in #99; this fixes a known error/warning when linter is run.


twiecki:
Thanks!

PR #105, opened 2019-04-15
[WIP]  Fix notebook issues on TF2

kyleabeauchamp:
1.  hmc.ipynb was referencing the old `_hmc` subpackage, which I've now pointed to `hmc`.
2.  `example_models.ipynb` was calling some tf1 functions, which I've re-directed towards the TF1/2 compat. layer.  This notebook is still broken, however, but
3.  In my version of jupyter (py37 + jupyter server 5.7.8), the %%time magic prevents variable definitions within the scoped cell.  This leads to inability of re-running the notebooks from start to finish.   I've commented out two magics as a workaround, which allows notebooks to be run start to finish without exception.  

I'm not sure what we want to do about (2), as my fix still leads to inability to run notebook.  If that notebook is considered "already broken", then we could merge this as-is with the goal of moving a bit closer to working.


kyleabeauchamp:
I'll split this into a "good" PR and a [WIP] PR for easier review.

PR #106, opened 2019-04-15
Fix HMC notebooks

kyleabeauchamp:
1.  hmc.ipynb was referencing the old `_hmc` subpackage, which I've now pointed to `hmc`.
2.  In my version of jupyter (py37 + jupyter server 5.7.8), the %%time magic prevents variable definitions within the scoped cell.  This leads to inability of re-running the notebooks from start to finish.   I've commented out the two magics as a workaround.


twiecki:
Thanks! What's your email?

kyleabeauchamp:
Same account name but at gmail

PR #107, opened 2019-04-15
[WIP] Add TF2 fix to notebook

kyleabeauchamp:
`example_models.ipynb` was calling some tf1 functions, which I've re-directed towards the TF1/2 compat. layer.  This change fixes the imports, but there's still something in cell 3 that leads to `TypeError: _DistributionMeta object got multiple values for keyword argument 'scale'`


kyleabeauchamp:
Is TF2 the target at this point?  I can't recall if a final decision was made...

twiecki:
Yes, TF2 is the target.

PR #111, opened 2019-05-28
Draft of working observe method

fonnesbeck:
`model.observe` now conditions on passed keyword argument data values; force `log_prob` to use keyword arguments.

### DIFF
--- a/pymc4/_model.py
+++ b/pymc4/_model.py
@@ -86,10 +86,10 @@ class Model:
         def log_prob(**kwargs):
             vars = self._forward_context.vars
             if set(self._observations.keys()).intersection(set(kwargs.keys())):
-                raise ValueError('double specified var')
+                raise ValueError('Passed variable already specified.')
             kwargs.update(self._observations)
             if len(kwargs) != len(vars):
-                raise ValueError('missing var')
+                raise ValueError('Missing value for variable in logp.')
             context = contexts.InferenceContext([kwargs[v.name] for v in self._forward_context.vars], expected_vars=self._forward_context.vars)
             with context:
                 self._evaluate()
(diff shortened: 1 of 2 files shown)


pymc4/_model.py
```diff
@@ -83,8 +83,14 @@ def _evaluate(self):
...
         """Return the log probability of the model."""
 
-        def log_prob(*args):
-            context = contexts.InferenceContext(args, expected_vars=self._forward_context.vars)
+        def log_prob(**kwargs):
+            vars = self._forward_context.vars
+            if set(self._observations.keys()).intersection(set(kwargs.keys())):
+                raise ValueError('double specified var')
+            kwargs.update(self._observations)
+            if len(kwargs) != len(vars):
```
canyon289: Out of scope for this PR but adding note. Having `pm.RandomVariable` objects referenced by key in a dictionary would help with checks in the `log_prob` function

https://github.com/pymc-devs/pymc4/issues/112

pymc4/_model.py
```diff
@@ -83,8 +83,14 @@ def _evaluate(self):
...
         """Return the log probability of the model."""
 
-        def log_prob(*args):
-            context = contexts.InferenceContext(args, expected_vars=self._forward_context.vars)
+        def log_prob(**kwargs):
+            vars = self._forward_context.vars
+            if set(self._observations.keys()).intersection(set(kwargs.keys())):
+                raise ValueError('Passed variable already specified.')
+            kwargs.update(self._observations)
+            if len(kwargs) != len(vars):
```
brianwa84: just realized we should probably check more than just len, otherwise len({'x','y'}) == len({'z','a'})

fonnesbeck: Yes, related to Ravin's comment above. If we moved from a list to a dictionary, we could compare keys.

fonnesbeck: Fixed in #118

PR #113, opened 2019-05-28
TST: Added the rugby model as a test case to the pymc4 tests

springcoil:
From the work at the summit.

I've written an example here and done a few sanity checks.


springcoil:
Some remarks, this is a stop gap before we get pm.Sample to work correctly.

There's some errors in the model specification. This will be fixed - cc @canyon289

springcoil:
@twiecki Do you think this needs to goto example models? Too large a model?

twiecki:
Yeah, maybe turn this into a NB and place it into the notebooks directory.

ferrine:
I will catch up this test case and port it taking in account new desing

ferrine:
BTW, is it actually a test case?

ericmjl:
@ferrine we discussed at the summit that having some simple examples placed in the test suite can be useful as an integration test, in case we make API changes.

canyon289:
@springcoil Did you want to update this to the new syntax or can we close? Just doing house cleaning

springcoil:
I've not got the time sadly.

ericmjl:
Closing this PR, as @ColCarroll has made the notebook.

PR #117, opened 2019-05-29
Improved variable checking when passing observed variables

fonnesbeck:
This improves the initial implementation of `observe` by checking variables by name.


fonnesbeck:
Bad rebase.

PR #121, opened 2019-05-29
[TST] Add test for hierarchical model indexing.

ericmjl:
In the same spirit as #113.

### DIFF
--- a/pymc4/tests/test_hierarchical_comparison.py
+++ b/pymc4/tests/test_hierarchical_comparison.py
@@ -1,0 +1,49 @@
+"""
+Test for hierarchical model.
+
+Intent behind test:
+*   Ensure that future API development yields the correct sample shape for
+    hierarchical models.
+"""
+import pymc4 as pm
+import tensorflow as tf
+import numpy as np
+import pytest
+import pandas as pd
+
+mapping = {"fortis": 0, "scandens": 1, "unknown": 2}
+
+df = pd.read_csv(
+    "https://raw.githubusercontent.com/ericmjl/bayesian-stats-modelling-tutorial/master/data/finch_beaks_2012.csv"
+)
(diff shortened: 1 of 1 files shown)


twiecki:
I think this would be better as an example in example_models. For a test better to have it small and concise.

springcoil:
@twiecki I'm not sure about this. I think it's good to have a few models in the tests. You worried about speed?

canyon289:
I'm worried about it not being an actual test. If we do this we should mark it as an integration test, or something other than a unit test

ericmjl:
@twiecki I actually agree with you, these are great examples to have, but at the same time, I think it's good to programmatically test a selective set of quick-running examples, whether in notebooks or not.

ferrine:
Do we consider it as a test case or it should be a notebook with an example?

twiecki:
That's the unanswered question. Is it useful to simple example models in the test suite. I think they make for weird tests and I'd prefer to just have them as example NBs, but don't have strong arguments.

ericmjl:
An alternative is to put it as an example that we programmatically execute each PR. That way we can be sure nothing breaks.

That said, it is more lightweight to edit a plain text file if the API changes.

My two cents on this matter :smile:.

canyon289:
@ericmjl  Did you want to update this to the new syntax or can we close? Just doing house cleaning

ericmjl:
@canyon289 I'd like to update this, but I'm having issues with tests failing locally in `master`, even after latest pull. Haven't had the time to debug though. When my bandwidth opens up, I'll raise an issue if problems still persist.

ColCarroll:
I problem I saw when setting up CI is that a library called `gast` gets installed at version 0.3.0, but tensorflow uses some features that are only available in 0.2.x. That _might_ help.

ericmjl:
I'm going to close this PR, as I've got different plans to start building some simpler examples into notebooks that show how to use the API.

PR #125, opened 2019-05-30
Generator models

ferrine:
The proposed model specification

```python
@pm.model(keep_auxiliary=False)
def Horseshoe(mu=0, tau=1., s=1.):
    scale = yield tfd.HalfCauchy(0, s, name="scale")
    noise = yield tfd.Normal(0, tau, name="noise")
    return scale * noise + mu

@pm.model
def linreg(x):
    scale = yield tfd.HalfCauchy(0, 1, name="scale")
    coefs = yield Horseshoe(tf.zeros(x.shape[1]), name="coefs")
    predictions = yield tfd.Normal(tf.linalg.matvec(x, coefs), scale, name="predictions")
    return predictions
```

TODO/goals:
- [x] pymc4 imports
- [ ] wrap tensorflow distributions
- [x] decide on transforms API
- [ ] decide on program transformations API (reparametrization). Can it be merged with transforms API?
- [x] implement logprob function as in old draft
- [x] try to keep implemenation backend agnostic
- [x] make it possible to convert backend distibutions to pymc4 distributions
- [x] make on demand autoimport for backends (some API?)

Neat Features
--------------
* Operator Overloading
instead of 

```python
@pm.model(keep_auxiliary=False)
def Horseshoe(mu=0, tau=1., s=1.):
    scale = yield pm.HalfCauchy(0, s, name="scale")
    noise = yield pm.Normal(0, tau, name="noise")
    return scale * noise + mu
```

You should be able to

```python
@pm.model(keep_auxiliary=False)
def Horseshoe(mu=0, tau=1., s=1.):
    noise = yield pm.Normal(0, tau, name="noise") * pm.HalfCauchy(0, s, name="scale")
    return noise + mu
```


pymc4/coroutine_model.py
```diff
@@ -0,0 +1,233 @@
+import biwrap
+import functools
+import types
+import pymc4
```
twiecki: better to use local imports

pymc4/distributions/discrete.py
```diff
@@ -58,12 +52,17 @@ class Bernoulli(RandomVariable):
...
     - p: probs
     """
+    def __init__(self, name, p, **kwargs):
+        super().__init__(name, p=p, **kwargs)
+
+    def lower_limit(self):
+        return 0
 
-    def _base_dist(self, p: TensorLike, *args, **kwargs):
-        return tfd.Bernoulli(probs=p, *args, **kwargs)
```
twiecki: Where do you link this to the tf distribution?

ferrine: I do not link yet, figuring out how to deal with transforms first

pymc4/flow/evaluate.py
```diff
@@ -0,0 +1,171 @@
...
+                .format(scoped_name))
+        if scoped_name in state.values:
+            return_value = state.values[scoped_name]
+        else:
+            return_value = state.values[scoped_name] = dist.sample()
+        state.distributions[scoped_name] = dist
+        return return_value, state
+
+    def prepare_model_control_flow(self, model, model_info, state):
+        control_flow: types.GeneratorType = model.control_flow()
```
junpenglao: This is pretty magically to me... what does this line do?

ferrine: model you see in arguments keeps a no-arg function that creates a generator that yields distributions. I name that generator `control_flow`. For some reason, my type checker was silly to detect generator type here, I helped it

junpenglao: Oh so the `: types.GeneratorType` part in `control_flow: types.GeneratorType = model.control_flow()` is type hint right?

ferrine: Right

rpgoldman: Minor grammar correction:

> "this may happen if you forget to use `pm.name_scope()` when calling same "

should be 

> "this may happen if you forget to use `pm.name_scope()` when calling the same "

fonnesbeck:
What did we decide regarding exposing the `yield` expressions? Are we going to hide them by wrapping TF distribtuions?

ferrine:
We found it a mistake. One is not possible to hide yield

pymc4/coroutine_model.py
```diff
@@ -0,0 +1,236 @@
+import biwrap
```
fonnesbeck: Can this module just be called `model.py`, or are we expecting other (non-coroutine) APIs for models?

pymc4/coroutine_model.py
```diff
@@ -0,0 +1,236 @@
...
+import functools
+import types
+import pymc4
+from pymc4.scopes import name_scope
+
+
+@biwrap.biwrap
```
fonnesbeck: Since biwrap is so small, can we just add it to utils rather than having an external dependency?

twiecki:
Can't we wrap them in functions that do the yield?

pymc4/coroutine_model.py
```diff
@@ -0,0 +1,236 @@
...
+            name = base_fn.__name__
+    return name
+
+
+class ModelTemplate(object):
+    """
+    The `ModelTemplate` object, this keeps a function, responsible for a specific generative process.
+    Generative process is one that samples from prior distributions and interacts
+    them in a user defined way arbitrarily complex.
+
```
fonnesbeck: Suggested re-wording: `ModelTemplate` is a callable object that represents a generative process. A generative process samples from prior distributions and allows them to interact in arbitrarily-complex, user-defined ways.

fonnesbeck:
Not even within `__new__`? I thought @twiecki sketched that out.

pymc4/coroutine_model.py
ferrine: Just tried to be explicit

ferrine:
No, unfortunately this was a my mistake to guess we can do that with new. Then I verified this is not possible:(

pymc4/coroutine_model.py
ferrine: Moreover we say any noarg generator function is a model template. I do not like default args and thinking over a more beautiful solution here

ferrine:
Answering @twiecki, no we can't, a generator function is a function that has a yield inside. Once we hide it from a generator function, this is not a generator function any more. The return value from new will be a generator

pymc4/coroutine_model.py
ferrine: We can

ferrine:
Ongoing goals (we better move this into a separate PR):
* wrap tensorflow distributions
* wrap tensorflow bijectors
* decide on program transformations API (reparametrization)

ferrine:
I did a rebase favouring my commits over master, so merge is going to be without conflicts. I'll devote some time to documentation tests and merge in an hour and merge once tests pass

PR #128, opened 2019-06-18
Tensorflow backend

sshekh:
Added Normal distribution as an example.

```python
import pymc4.distributions as dist

n = dist.Normal(mu=0, sigma=1, name='n')
n.log_prob(3)   #-5.4189386
n.sample(5)  #returns 1-D numpy array of length 5
```

* Every distribution has an `_init_backend()` which does backend specific initialization. This way the wrapper doesn't need to override `__init__` for every backend
*  Assumes eager mode. tf-graph is not supported so far
* API is similar to tf-probability (`log_prob` and `sample`). Although this can be changed to be closer to PyMC3

* Every backend will have its own set of wrappers in `distributions/` ~~and its own `AbstractBackend` derived class.~~. EDIT: AbstractBackend has been removed for now. It can be easily added later.
* Every backend will have its own `BackendDistribution` class

TODO/goals
- [x] Backend is currently hardcoded  in `distributions/backend.py`. Add proper way to switch backends. EDIT: using `PYMC_BACKEND` environment variable like in Keras.
- [ ] Maintain state
- [x] Better mapping of arguments/keywords between PyMC4 and backends. Maybe use a dictionary to switch between the parameter mappings. EDIT: This has been implemented using `conditions` dictionary.
- [ ] Add more distributions

### DIFF
--- a/pymc4/distributions/__init__.py
+++ b/pymc4/distributions/__init__.py
@@ -1,1 +1,8 @@
-from . import base
+__backend = 'tensorflow'
+
+if __backend == 'base':
+    from .base import *
+elif __backend == 'tensorflow':
+    from .tensorflow import *
+else:
+    raise Exception('Backend %s not supported' % __backend)
--- a/pymc4/distributions/__init__.py
+++ b/pymc4/distributions/__init__.py
@@ -1,8 +1,10 @@
-__backend = 'tensorflow'
+import os
 
-if __backend == 'base':
-    from .base import *
-elif __backend == 'tensorflow':
+_backend = 'tensorflow'
+if 'PYMC_BACKEND' in os.environ:
+    _backend = os.environ['PYMC_BACKEND']
+
+if _backend == 'tensorflow':
     from .tensorflow import *
 else:
     raise Exception('Backend %s not supported' % __backend)
--- a/pymc4/distributions/__init__.py
+++ b/pymc4/distributions/__init__.py
@@ -7,4 +7,4 @@ if 'PYMC_BACKEND' in os.environ:
 if _backend == 'tensorflow':
     from .tensorflow import *
 else:
-    raise Exception('Backend %s not supported' % __backend)
+    raise Exception('Backend %s not supported' % _backend)
--- a/pymc4/distributions/tensorflow/continuous.py
+++ b/pymc4/distributions/tensorflow/continuous.py
@@ -8,68 +8,14 @@ import tensorflow_probability as tfp
 tfd = tfp.distributions
 
 
-from .distribution import ContinuousDistribution
+from .distribution import BackendDistribution
+from ..base.continuous import Normal
 
 __all__ = ['Normal']
 
 
-class Normal(ContinuousDistribution):
+class Normal(BackendDistribution, Normal):
     r"""
-    Univariate normal random variable.
-
-    The pdf of this distribution is
-
-    .. math::
-
-       f(x \mid \mu, \tau) =
-           \sqrt{\frac{\tau}{2\pi}}
-           \exp\left\{ -\frac{\tau}{2} (x-\mu)^2 \right\}
-
-    Normal distribution can be parameterized either in terms of precision
-    or standard deviation. The link between the two parametrizations is
-    given by
-
-    .. math::
-
-       \tau = \dfrac{1}{\sigma^2}
-
-    .. plot::
-
-        import matplotlib.pyplot as plt
-        import numpy as np
-        import scipy.stats as st
-        plt.style.use('seaborn-darkgrid')
-        x = np.linspace(-5, 5, 1000)
-        mus = [0., 0., 0., -2.]
-        sigmas = [0.4, 1., 2., 0.4]
-        for mu, sigma in zip(mus, sigmas):
-            pdf = st.norm.pdf(x, mu, sigma)
-            plt.plot(x, pdf, label=r'$\mu$ = {}, $\sigma$ = {}'.format(mu, sigma))
-        plt.xlabel('x', fontsize=12)
-        plt.ylabel('f(x)', fontsize=12)
-        plt.legend(loc=1)
-        plt.show()
-
-    ========  ==========================================
-    Support   :math:`x \in \mathbb{R}`
-    Mean      :math:`\mu`
-    Variance  :math:`\dfrac{1}{\tau}` or :math:`\sigma^2`
-    ========  ==========================================
-
-    Parameters
-    ----------
-    mu : float
-        Mean.
-    sigma : float
-        Standard deviation (sigma > 0).
-
-    Examples
-    --------
-    .. code-block:: python
-        @pm.model
-        def model():
-            x = pm.Normal('x', mu=0, sigma=10)
-
     Developer Notes
     ---------------
     Parameter mappings to TensorFlow Probability are as follows:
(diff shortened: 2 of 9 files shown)


pymc4/distributions/__init__.py
```diff
@@ -1 +1,8 @@
-from . import base
+__backend = 'tensorflow'
```
ferrine: We can leave it now as is or use Keras approach:
https://github.com/keras-team/keras/blob/master/keras/backend/load_backend.py#L75

pymc4/distributions/__init__.py
```diff
@@ -1 +1,8 @@
...
+
+if __backend == 'base':
```
ferrine: Base backend is not runnable, I see no case when we might need to load it

pymc4/distributions/tensorflow/continuous.py
```diff
@@ -0,0 +1,97 @@
...
+import tensorflow_probability as tfp
+tfd = tfp.distributions
+
+
+from .distribution import ContinuousDistribution
+
+__all__ = ['Normal']
+
+
+class Normal(ContinuousDistribution):
```
ferrine: We can try to avoid copy-pasting docs, probably with InheritDocs metaclass

ferrine: or this simpler solution https://stackoverflow.com/a/2025599

pymc4/distributions/tensorflow/continuous.py
```diff
@@ -0,0 +1,97 @@
...
+        self.sigma = sigma
+        super().__init__(name, **kwargs)
+        self._init_backend()
+
+    def _init_backend(self):
+        self._backend_dist = tfd.Normal(
+            loc=self.mu, scale=self.sigma, name=self.name)
+
+    def sample(self, shape=(), seed=None):
+        return self._backend_dist.sample(shape, seed).numpy()
```
ferrine: That should not be converted to Numpy as we will use this within the backend. Instead, we can add `sample_numpy` that does all for conversion (should be possible in base class).

pymc4/distributions/tensorflow/continuous.py
```diff
@@ -0,0 +1,97 @@
...
+    Developer Notes
+    ---------------
+    Parameter mappings to TensorFlow Probability are as follows:
+
+    - mu: loc
+    - sigma: scale
+    """
+
+    def __init__(self, name, mu, sigma, **kwargs):
+        self.mu = mu
```
ferrine: the previous `conditions` dictionary might allow you to only override `_init_backend` in inheritance, where you access members of `conditions`. And the only place you construct this dictionary is a Normal distribution without backend

```python
class Normal(Distribution):
    def __init__(self, mu, sigma, ...):
        ...


class BackendNormal(Normal):
    def _init_backend(self):
        mu, sigma = self.conditions["mu"], self.conditions["sigma"]
        ...

```

pymc4/distributions/tensorflow/continuous.py
ferrine: Better move sample and log_prob into a base class for backended distributions
```python
class BackendDistribution(Distribution):
    def sample(self, ...):
        ...
    def log_prob(self, ...):
        ...

class BackendNormal(BackendDistribution, Normal):
    def _init_backend(self, ...):
        ...
```
This compositional inheritance will affect the method resolution order. If in eg you override `__init__` in BackendDistribution (this should not be the case), it will be resolved to there, but as you only override `sample`, `logprob` to interact with backend distribution and `_init_backend` is defined right here, you resolve

* `__init__` -> Normal
* `_init_backend` -> BackendNormal
* `sample`, etc -> BackendDistribution

The formal method resolution order is gonna be like this `BackendNormal->BackendDistribution->Normal->Distribution->Model`

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,66 @@
...
+                 transform=None,
+                 **kwargs):
+        super().__init__(
+            self.unpack_distribution,
+            name=name,
+            keep_auxiliary=keep_auxiliary,
+            keep_return=keep_return,
+        )
+        self._init_backend()
+        pass
```
ferrine: [style] no need in `pass` here

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def unpack_distribution(self):
+        return unpack(self)
+
+    @abc.abstractmethod
+    def _init_backend(self):
+        pass
+
+    @abc.abstractmethod
+    def sample(self, shape=(), seed=None):
```
ferrine: As mentioned above `self._backend_distribution.sample(*args, **kwargs)` shoul work in 90% cases

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,66 @@
...
+        seed : int|None
+            random seed
+        Returns
+        ---------
+        array
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def log_prob(self, value):
```
ferrine: same comment as above

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,66 @@
...
+        raise NotImplementedError
+
+
+class ContinuousDistribution(Distribution):
```
ferrine: No need in defining Continuous class here. `isisinstance(backend_dist_normal, base.ContinuousDistribution)` should work as not backended `Normal` inherits from base.ContinuousDistribution

pymc4/distributions/__init__.py
sshekh: Added enviroment checking option. Although only tensorflow backend is supported for now.

pymc4/distributions/__init__.py
sshekh: removed.

pymc4/distributions/tensorflow/continuous.py
sshekh: Cool solution! Reduces a lot of copy-paste in class docstrings. I implemented the simpler verision
```
__doc__ = Distribution.__doc__ + __doc__
```
Although it seems to work only for class docstrings, not for individual methods. Also requires that base class docstring is not None.

pymc4/distributions/tensorflow/continuous.py
sshekh: `BackendDistribution` added and `sample` and `log_prob` moved there. You are right, it should cover most of the cases. In the case when it doesn't, we can simply override it.

pymc4/distributions/tensorflow/continuous.py
sshekh: Done. So `BackendNormal` (implemented as `tensorflow.Normal`) overrides only `_init_backend()`. The `conditions` dictionary is already constructed in `Distribution` so I don't think `base.Normal` will need to override any method at all (Currently it just overrides `__init__`)

pymc4/distributions/tensorflow/distribution.py
sshekh: done

pymc4/distributions/tensorflow/distribution.py
sshekh: Nice suggestions! Every backend only needs to define `BackendDistribution` now

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,23 @@
...
+        ---------
+        Tensor
+        """
+        return self._backend_distribution.sample(*args, **kwargs)
+
+    def sample_numpy(self, *args, **kwargs):
+        """Return raw numpy format of sampling"""
+        return self.sample(*args, **kwargs).numpy()
+
+    def log_prob(self, value):
```
ferrine: `log_prob_numpy` is missing. Better add this to the Statistical Distribution class and override here

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def sample(self, *args, **kwargs):
+        """
+        Returns
+        ---------
+        Tensor
+        """
+        return self._backend_distribution.sample(*args, **kwargs)
+
+    def sample_numpy(self, *args, **kwargs):
```
ferrine: better keep arguments explicit, there may be more of them

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class BackendDistribution(Distribution):
+    """
+    BackendDistribution for Tensorflow distributions
+    """
+    __doc__ = Distribution.__doc__ + __doc__
+
+    def sample(self, *args, **kwargs):
+        """
```
ferrine: that's why you may miss the docstring. If you leave it empty, it will be filled from inheritance

PR #129, opened 2019-06-24
Yield transforms

ferrine:
Adding a model modificator to handle transforms. Basic Idea is to work with potentials and transforms only, without referring to a backend. What is done in `modify_distribution` is working directly in the transformed space. Total logp is augmented with potentials.


ferrine:
the notebook I was able to play with
https://gist.github.com/ferrine/80e78c09222d527281dba7b6e7d7a415

pymc4/distributions/tensorflow/transforms.py
```diff
@@ -0,0 +1,17 @@
+import tensorflow as tf
+from .. import abstract
+
+
+class Log(abstract.transforms.Log):
```
twiecki: Why can't we use the TFP bijectors?

ferrine: We can, but I found no reason to do that here

pymc4/flow/evaluate.py
```diff
@@ -180,3 +182,76 @@ def finalize_control_flow(self, stop_iteration, model_info, state):
...
+class TransformedSamplingExecutor(SamplingExecutor):
+    def modify_distribution(self, dist, model_info, state):
+        dist = super().modify_distribution(dist, model_info, state)
+        if not isinstance(dist, pymc4.distributions.abstract.Distribution):
+            return dist
+        # this will be mentioned later, that's the way to avoid loops
+        if dist.transform is None:
+            return dist
+        transform = dist.transform
+        scoped_name = pymc4.scopes.Scope.variable_name(dist.name)
```
twiecki: Use relative names.

ferrine:
I've restructured code and since executor is stateless (unless inited with particular state), I've added default instances of executors under the `pm.flow` namespace

ferrine:
an updated version of trial version with transforms is here
https://gist.github.com/ferrine/8937bb5905651873970f4d8af1d28baa

ferrine:
I also renamed a couple of modules, new names seemed to me more semantically appropriate

pymc4/distributions/tensorflow/transforms.py
twiecki: Avoiding code duplication and thus bugs.

pymc4/flow/executor.py
```diff
@@ -118,29 +123,31 @@ def evaluate_model(self, model, state=None):
...
+    __call__ = evaluate_model
+
 
 SamplingState = collections.namedtuple("EvaluationState", "values,distributions,potentials")
 
 
 class SamplingExecutor(Executor):
-    def __init__(self, **do):
-        self.default_do = do
+    def __init__(self, *do: dict, **do_kwargs: dict):
```
twiecki: what does `do` mean? the `do` operator, conditioning?

pymc4/flow/transformed_executor.py
```diff
@@ -0,0 +1,79 @@
+import copy
+
+from pymc4 import scopes, distributions
+from pymc4.distributions import abstract
+from pymc4.distributions.abstract.transforms import JacobianPreference
+from pymc4.flow import SamplingExecutor
+
+
+class TransformedSamplingExecutor(SamplingExecutor):
+    def modify_distribution(self, dist, model_info, state):
```
twiecki: This code here is pretty hard to follow and seems rather complex, sure this is the right design?

pymc4/flow/executor.py
ferrine: Exactly

pymc4/flow/transformed_executor.py
ferrine: The source of complexity comes from the following desirable use cases:

1. You want to get the initial point in the transformed space. Here you usually provide test values in the untransformed space and expect transform to be applied. That is what in
https://github.com/pymc-devs/pymc4/pull/129/files/5ec9546526c26ef7eac454f43bcc48951b4d479e#diff-38d8573a4755732b98ab055cf77822adR29
done

```
given: untransformed
do:
    transform
    record transformed
    record logdet
    return untransformed
end
```

2. You actually want to sample and provide values in the transformed space
https://github.com/pymc-devs/pymc4/pull/129/files/5ec9546526c26ef7eac454f43bcc48951b4d479e#diff-38d8573a4755732b98ab055cf77822adR23
Probably we would exclude the case when both values (in the transformed and untransformed case are passed). But it is proceeded silently for now. Maybe warning?
```
given: transformed
do:
    inverse transform
    record untransformed
    record logdet
    return untransformed
end
```

3. You want to sample in the transformed space, the same purpose as in 1., to get initial values for a chain.

```
given: nothing
do:
    sample untransformed 
    # ^--here we are supposed to yield or manually call .sample()
    # I decided to call yield as additional safety checks are performed
    record untransformed
    transform
    record transformed
    record logdet
    return untransformed
end
```

The most tricky part is probably sampling, you can't just sample, you need to record it into the state dict after transform applied. But you first apply the transform and expect the user to get the return value. So sampling comes first, transform second. You see, that differs from 1 and 2 case, when we did no sampling at all.

ferrine: In pymc3 we had a concept of transformed distributions
https://github.com/pymc-devs/pymc3/blob/master/pymc3/model.py#L1557
We manually applied one to the random variable. Do we need the same and how does this look like in here for the same purpose? The very first challenge I got thinking about it is how to decide on the 3 cases described above. Finally you may come up with mixed transformed and untransformed distributions in the same state, but the difference is only `logdet` between them. You should additionally think about conventions "when to keep one or another" and "how to check for duplicates". After a clear picture was not obtained, I decided on a dumb idea. If you want to compute logp for the transformed space, you'll see logdets added, the rest computations and rules are the same.

pymc4/distributions/tensorflow/transforms.py
ferrine: There is some ambiguity with tfp bijectors. sometimes the do require event_dims, sometimes not:
https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/bijectors/bijector.py#L1217
https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/bijectors/power_transform.py#L94

pymc4/flow/executor.py
twiecki: I'm not sure that's right here, do != conditioning as I understand it. do means if I were to fix a value and the causal links work correctly (i.e. fixing the thermometer to a value does not change conditional of the temperature). Plain conditioning does not necessarily incorporate causality (i.e. the conditional distribution of the temperature is dependent on the thermometer value). I would just call it conditioning here, maybe `fix`, `cond`, `condition`, `depends`, while I appreciate the succinctness of `do`.

pymc4/flow/transformed_executor.py
twiecki: Thanks for the detailed explanation, unfortunately I'll be traveling the rest of the week and not sure I can read this in detail soon, but do remind me next week if I don't get back to you. Or maybe @ColCarroll and @junpenglao have thoughts.

pymc4/flow/executor.py
ferrine: The way it appears in the graph, this is exactly a `do` operator. In case you have the following dependence and state provided for 
```
do = {"a": 1}
graph:
    b->a->c
```
You come up with
```
    b->do(a, inserted from state)->c
```

ferrine: However, for the leaf nodes, the observed, do=conditioning

ferrine: My fault, the logp computation seems to work as conditioning, you are right

ferrine:
I've added a bit more cleaner comments and changed the signature to `conditions` instead of `do
`

ferrine:
Is everyone fine with the PR and I merge it?

twiecki:
@ferrine I really think we should be using the TFP bijectors, it makes no sense to have duplication of things like stick-breaking etc which are quite tricky to get right. If they have shape issues we should work with upstream to fix them.

ferrine:
Agreed, as for this particular PR I would suggest to leave this as a disign direction and address the bijectors in further PRs. The reason is this PR is slowing down the process on distributions. The way to work on them so far is on top of the old branch, but you will have to merge this PR at some point that makes it less effective.

ferrine:
What I expect from the final review is to agree or disagree with the proposed design choice. At a high level it is:
* Proceed transforms within a separate executor. Motivation is that model is different in transformed and untransformed space (e.g. logp)
* logp computation is a part of the state that appears after each model forward evaluation. Motivation is to make things modular and self-complete: you should not figure out about state structure anywhere. 
* changed names of directories, probably a bad idea to do it right now, but I found them too misleading

Things I propose to leave for my future PR:
* tighter integration of pymc4 transforms and tfp bijectors. I observed some shape complexity and given little time did not find an elegant solution. It requires reading sources more closely than I briefly did.
*

fonnesbeck:
Is there any reason for this not to be in master? Seems strange to be managing pull requests for another branch,

ferrine:
@fonnesbeck no tests and non stable design was a stop factor for me. I will polish the yield branch today and merge it into master

PR #130, opened 2019-07-04
[WIP] Adding more distributions

sshekh:
Will be adding more distributions (continuous as well as discrete) and tests here. Plan to use helper functions and classes as in [pymc3/test_distributions.py](https://github.com/pymc-devs/pymc3/blob/master/pymc3/tests/test_distributions.py)


pymc4/distributions/tensorflow/continuous.py
```diff
@@ -11,7 +11,62 @@
...
+    Developer Notes
+    ---------------
+    Parameter mappings to TensorFlow Probability are as follows:
+
+    - alpha: concentration1
+    - beta: concentration0
+    """.format(
+        abstract.Beta.__doc__
+    )
+    __doc__ = abstract.Beta.__doc__ + __doc__
```
ferrine: Duplicated docstring

pymc4/distributions/tensorflow/continuous.py
```diff
@@ -11,7 +11,62 @@
 
 tfd = tfp.distributions
 
-__all__ = ["Normal", "HalfNormal"]
+__all__ = ["Beta", "Cauchy", "HalfCauchy", "Normal", "HalfNormal"]
+
+class Beta(BackendDistribution, abstract.Beta):
+    __doc__ = r"""{}
```
ferrine: New line is expected

tests/test_distributions.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+import math
+import numpy as np
+import random
+import scipy.stats as sts
+from pymc4 import distributions as dist
+from numpy.testing import assert_almost_equal
+
+def test_normal():
```
twiecki: Not sure we need these tests because they already happen in TFP, instead we should test that our parameterization matches what TFP provides.

ferrine: We need a test to match scipy parametrization since the implementation API is backend agnostic.  We then just change env var in Travis to test other backends

ferrine: I would test the change of parameters additionally

sshekh: @twiecki  so I was trying to add tests just like in [pymc3/test_distributions.py](https://github.com/pymc-devs/pymc3/blob/master/pymc3/tests/test_distributions.py) i.e using `Domains`, `Point()` etc. However @ferrine  suggested to look into the [hypothesis](https://hypothesis.readthedocs.io/en/latest/details.html) library for testing distributions. 

I will look into the library (it's supposed to integrate well with pytest) and make a draft PR of how the tests are supposed to look. Just adding discrete distributions for now. Thank you for your patience.

ferrine:
I've changed the base branch to the `observed` branch since we point these changes there. It should simplify the review omitting commits from `observed`

pymc4/distributions/abstract/discrete.py
```diff
@@ -220,18 +206,16 @@ class Categorical(BoundedDiscreteDistribution):
...
 
     def __init__(self, name, p, **kwargs):
         super().__init__(name, p=p, **kwargs)
 
-    # TODO: upper limit
+    def lower_limit(self):
+        return 0
+
+    def upper_limit(self):
+        return len(self.conditions["p"]) - 1
```
ferrine: this might behave inconsistently with broadcasting. We can make this method to be backend specific if generic `self.conditions["p"].shape[-1]` fails for some backend
```suggestion
        return self.conditions["p"].shape[-1] - 1
```

lucianopaz:
This PR seems to be outdated. We've added some distributions but made other major API changes. I'll close this but feel free to open a new PR of you are interested in adding some distributions that are currently missing.

PR #131, opened 2019-07-14
Observed Variables

ferrine:
Important design changes

* `SamplingState` class has been changed to distinguish observed from regular variables
* added `Dist(..., observed=value)` API
* added a protocol for suppressing values provided in `Dist(..., observed=value)` mapping the name to None in observed dict. (Un)transformed values are only can be set instead of observed if the observed is suppressed.
* added `Dist.dist(...)` api that allows creating distributions that yet do not have a name. To use the created distribution one should call `dist.prior(name)` on it.

### DIFF
--- a/pymc4/flow/executor.py
+++ b/pymc4/flow/executor.py
@@ -230,9 +230,15 @@ class SamplingExecutor(object):
     """
 
     def validate_return_object(self, return_object: Any):
-        if isinstance(return_object, coroutine_model.Model):
+        if isinstance(return_object, (coroutine_model.Model, types.GeneratorType, abstract.Potential)):
             raise EvaluationError(
-                "Return values should not contain instances of pm.coroutine_model.Model"
+                "Return values should not contain instances of "
+                "apm.coroutine_model.Model`, "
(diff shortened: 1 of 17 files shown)


pymc4/flow/executor.py
```diff
@@ -33,35 +54,315 @@ class EarlyReturn(StopIteration):
...
+        self.untransformed_values = untransformed_values
+        self.observed_values = observed_values
+        self.all_values = collections.ChainMap(
+            self.untransformed_values, self.transformed_values, self.observed_values
+        )
+        self.distributions = distributions
+        self.potentials = potentials
+
+    def collect_log_prob(self):
+        return sum(
```
ferrine: I'm concerned about this part. It works for scalar values, but for vector valued log_prob this will fail. We need to access backend ops to correctly perform reduce operation.

canyon289: This comment is on oudated line. Is it still a concern of yours?

ferrine: Outdated

PR #134, opened 2019-07-22
Change random seed to session

canyon289:
I dont think we need to set the random seed for each test.

Admittedly though I also don't know how the global state of TF interacts with each test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,6 @@ import pytest
 import tensorflow as tf
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="session")
 def tf_seed():
     tf.random.set_seed(37208)  # random.org


ColCarroll:
I would actually leave this as a function level scope, but I welcome your disagreement:

1. The seed is fast to set, so this does not introduce much overhead
2. This makes tests non-deterministic in the presence of other tests. As a contrived example, this first "test suite" passes, with the seed set on each function, but the second fails:

```python
# Test 1
np.random.seed(0)
assert np.random.normal(size=1000).mean() < 0

# Test 2
np.random.seed(0)
assert np.random.normal(size=1000).mean() < 0
```

But this "test suite" fails:
```python
# Session seed
np.random.seed(0)

# Test 1
assert np.random.normal(size=1000).mean() < 0

# Test 2
assert np.random.normal(size=1000).mean() < 0
```

I think writing a good test suite will be very hard without gaming the random seeds at least a little, but I know you have thought more about this!

PR #135, opened 2019-08-03
Notebook cleanup

canyon289:
House cleaning for notebooks that no longer work and also are confusing since we've switched design.

Also added @ferrine design guide found at this gist 
https://gist.github.com/ferrine/59a63c738e03911eacba515b5be904ad


ferrine:
Do you plan adding any more changes? Can I merge?

fonnesbeck:
If they are non-functional, should we just remove them? They can always be recovered from Git if needed.

canyon289:
I do not plan on adding anymore changes. Everything in context_design is non functional based on generator design.

My only argument against removing them is that for now they still have some value to me in seeing what things used to look like for a reference but like you said we can certainly recover them for git.

PR #136, opened 2019-08-20
Sampling

ferrine:
# Contents

In this PR I'm going to implement a minimal implementation required to perform NUTS sampling backed by tensorflow.

The workflow is as follows:

1. Decide on API (input/output/namings)
    * initialization
    * calling sampling
    * accepted types
    * output format

2. Implement the rest based on design choice
    * in progress

# Discussion
### Initialization
I propose to allow override observed values only and expose `observed` kwarg. The rest tweaks are quite advanced to think of them right now. Don't want to make initialization that complicated for now.

The reference implementation for that may be found in `pymc4.inference.utils.initialize_state` function. All is done in there is basically a forward pass of the model and dropping unneeded stuff.

Things that can be improved here, but not in this PR:
* Set a test point manually (+ ability to pass an observed along with it as mentioned by @fonnesbeck)
* More advanced staff for model manipulation. E.g. removing and adding observed.

### Calling sampling
Sampling accepts sampler specific arguments along with initialization ones. 

Things that can be improved here, but not in this PR:
* Add more samplers
* Better abstractions from the backend

### Accepted types
For now, I limit sampling support only for wrapped generator functions with `pm.model`. The main purpose of doing that is flexibility in reusing the same model over and over.

Things that can be improved here, but not in this PR:
* Support plain generators. (no sure we need that)

### Output format
In pymc3 we had plenty of trace backends. The main requirement I see crucial for making things work is `arvis` integration. So, for now, I prefer not to add complexity adding multiple trace backends but simply return a dictionary with numpy arrays.

Things that can be improved here, but not in this PR:
* Multiple trace backends


pymc4/inference/__init__.py
```diff
@@ -0,0 +1,6 @@
+from pymc4 import _backend
+
+if _backend.TENSORFLOW:
```
fonnesbeck: I thought we were ditching the idea of multiple backends?

pymc4/inference/tensorflow/sampling.py
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+    if not isinstance(model, Model):
+        raise TypeError(
+            "`sample` function only supports `pymc4.Model` objects, but you've passed `{}`".format(
+                type(model)
+            )
+        )
+    if state is not None and observed is not None:
+        raise ValueError("Can't use both `state` and `observed` arguments")
```
fonnesbeck: Wouldn't you be passing initial values to `state` and data to `observed`?

pymc4/inference/__init__.py
ferrine: Not actually, we ditch the idea to put additional abstractions for now and treat every backend separately. So far we have the only one. Once we come up with a new backend we may try to implement it.

pymc4/inference/tensorflow/sampling.py
ferrine: Hmm, sounds reasonable. For now, I only have `initialize_state` util that does the stuff...

ferrine: Looking and the current implementation I got a bit stuck. I hope it is ok to mark this proposal as a desirable optional one for the future PRs.

pymc4/distributions/tensorflow/distribution.py
```diff
@@ -7,6 +7,12 @@ class BackendDistribution(Distribution):
 
     _backend_distribution: tfd.Distribution = None  # make type checkers happy
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.plate is not None:
```
ferrine: Hmmm, this might work for us

pymc4/distributions/tensorflow/transforms.py
```diff
@@ -1,6 +1,8 @@
 import tensorflow as tf
 from pymc4.distributions import abstract
 
+__all__ = ["Log"]
+
 
 class Log(abstract.transforms.Log):
```
twiecki: We should really replace these with TFP bijectors.

pymc4/inference/tensorflow/sampling.py
```diff
@@ -88,34 +89,42 @@ def sample(
 
     """
     logpfn, init = build_logp_function(model, state=state, observed=observed)
-    init = [tf.tile(tf.expand_dims(tens, 0), [num_chains] + [1] * tens.ndim) for tens in init]
+    init = [
+        tf.tile(tf.expand_dims(tens, 0), [num_chains] + [1] * tens.ndim)
+        for tens in init
+    ]
+
+    def parallel_logpfn(*state):
```
junpenglao: Could we isolate this function to Model as a property? it is otherwise difficult to test.

PR #138, opened 2019-09-03
Hacks to get vmap and XLA to work

twiecki:
OK, I got vmap and XLA to work, but not together. The NB shows things in action.

This highlights the current problems we have in the code-base but is not supposed to solve them in a clean way. I just hacked things to get them to work. Specifically:
 * `vmap` seems to wrap the arguments in a tuple which I just unpack here.
 * vmap seems to not like our custom transformers. I *strongly* advise to use TFP's bijectors as I do here.
 * XLA fails with `AttributeError: 'IndexedSlices' object has no attribute 'set_shape'` which I think the TFP guys saw before
 * `tf.function` compiling the sampling run itself fails because the executor *always* does a forward sample pass, even if it's not required like when evaluating the logp. compilation does not like sampling
 * We can't currently have `pm4.Normal('mu', 0, 1)` but need to convert them to scalar tensors, not sure why but vmap fails otherwise.


ferrine:
I would better match the pymc4 bijectors API against tfp, this will ease the integration and allow using tfp bijectors explicitly (but I have some conserns about that)

twiecki:
@ferrine That's fine, just not to recode them. I only remember you mentioning some shape inconsistency in one place, what other reasons are there?

ferrine:
My other concern is a default transform. You may find my current workaround in abstract constrained distributions. This is not the complete implementation and covers only positive values and may change (that is the concern). However, at this state it will bypass TF transform as is.

twiecki:
@ferrine I don't see how that argument relates to where the transforms come from - either hand-crafted or TFP bijectors.

pymc4/distributions/tensorflow/transforms.py
```diff
@@ -1,24 +1,26 @@
...
 class Log(abstract.transforms.Log):
-    # do not use tfp bijectors if transform is not too complicated
+    def __init__(self):
+        # NOTE: We actually need the inverse to match PyMC3, do we?
+        self._backend_transform = tfb.Exp()
+
     def forward(self, x):
-        return tfb.Exp().forward(x)
-        #return tf.math.log(x)
+        return self._backend_transform.inverse(x)
```
twiecki: Shouldn't we push this into the abstract class?

PR #145, opened 2019-09-10
Lint tests

canyon289:
Add linting/black/docstyle checks for tests and add pip freeze for package versions


ColCarroll:
thanks, this is nice!

PR #147, opened 2019-09-12
Add deterministic variable

ColCarroll:
I also removed the `tf_seed` fixture, since it is autoused.

The test suite shows how deterministics work right now:

```python
@pm.model
def half_normal(mu, sigma):
    normal = yield pm.Normal(name='normal', mu=mu, sigma=sigma, plate=3)
    yield pm.Deterministic(name="abs_normal", value=tf.math.abs(normal))

_, state = pm.evaluate_model(half_normal(0, 1))
print(state.untransformed_values['half_normal/normal'].numpy(),
      state.untransformed_values['half_normal/abs_normal'].numpy())
```
gives something like
```python
(array([-0.02932515, -1.9230154 ,  0.9288907 ], dtype=float32),
 array([0.02932515, 1.9230154 , 0.9288907 ], dtype=float32))
```

I'll wait on merging this, since I am not sure if I missed something in the design.

### DIFF
--- a/pymc4/distributions/abstract/distribution.py
+++ b/pymc4/distributions/abstract/distribution.py
@@ -203,3 +203,22 @@ class Potential(object):
     @abc.abstractmethod
     def value_numpy(self):
         raise NotImplementedError
+
+
+class Deterministic(Model):
+    """An object that can be sampled, but has no log probability."""
+
+    def __init__(self, name: Optional[NameType], value):
+        super().__init__(
+            self.unpack_distribution, name=name, keep_return=True, keep_auxiliary=False
+        )
+        self._init_backend()
+        self.value = value
+
+    def unpack_distribution(self):
+        return unpack(self.value)
+
+    @abc.abstractmethod
+    def _init_backend(self):
+        """Initialize the backend."""
+        pass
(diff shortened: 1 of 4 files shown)


ColCarroll:
I see what you mean -- sampling works, but just ignores this node.

ferrine:
Yea, sampling gonna be a problem, I don't know the solution yet

junpenglao:
@ferrine - can we ignore this node in sampling, and do a one pass forward transformation once sampling is done to log the deterministic?

ferrine:
@junpenglao, i am not sure about how to treat deterministic node yet. I would ask tfp side. It was in our wishlist on last call.

ferrine:
It is not about one forward pass. You should repeat it for all the samples

junpenglao:
> It is not about one forward pass. You should repeat it for all the samples

We can batch the function with `pfor` and compute for all samples at once.

ferrine:
Yep, but all the computations are done twise. This is not what i would expect

junpenglao:
> Yep, but all the computations are done twice. This is not what i would expect

Agree. I guess in general we should expect additional logging to be cheaper than running the computation twice.

kyleabeauchamp:
Looks very useful (e.g., was just wanting this feature), I'm still not caught up enough on the backend side to fix the sampler issue though.

ColCarroll:
Not needed after #161

PR #148, opened 2019-10-03
Add documentation and tests for biwrap

canyon289:
In this PR, and perhaps a couple more, I ask a bunch of questions about what code does, then add that to docs so hopefully everyone knows what it does

### DIFF
--- a/pymc4/coroutine_model.py
+++ b/pymc4/coroutine_model.py
@@ -66,13 +66,13 @@ class ModelTemplate:
     Parameters
     ----------
     template : callable
-        generative process, that accepts any arguments as conditioners and returns realizations if any.
+        Generative process, that accepts any arguments as conditioners and returns realizations if any.
     keep_auxiliary : bool
-        generative process may require some auxiliary variables to be created, but they are probably will not be used
+        Generative process may require some auxiliary variables to be created, but they are probably will not be used
         anywhere else. In that case it is useful to tell PyMC4 engine that we can get rid of auxiliary variables
         as long as they are not needed any more.
     keep_return : bool
-        the return value of the model will be recorded
+        The return value of the model will be recorded
     """
 
     def __init__(self, template, *, name=None, keep_auxiliary=True, keep_return=True):
@@ -160,6 +160,7 @@ class ModelTemplate:
             keep_auxiliary = self.keep_auxiliary
         if keep_return is None:
             keep_return = self.keep_return
+
         return Model(genfn, name=name, keep_auxiliary=keep_auxiliary, keep_return=keep_return)
 
 
--- a/pymc4/utils.py
+++ b/pymc4/utils.py
@@ -56,7 +56,7 @@ def merge_dicts(*dicts: dict, **kwargs: dict):
 
 
 def biwrap(wrapper):
-    """Decorator that allows for optional keyword arguments in lower level decoratrors.
+    """Allow for optional keyword arguments in lower level decoratrors.
 
     Notes
     -----
--- a/pymc4/utils.py
+++ b/pymc4/utils.py
@@ -64,6 +64,7 @@ def biwrap(wrapper):
     keep_auxiliary and keep_return. See pm.Model for all possible keyword parameters
 
     """
+
     @functools.wraps(wrapper)
     def enhanced(*args, **kwargs):
 
--- a/pymc4/utils.py
+++ b/pymc4/utils.py
@@ -58,6 +58,8 @@ def merge_dicts(*dicts: dict, **kwargs: dict):
 def biwrap(wrapper):
     @functools.wraps(wrapper)
     def enhanced(*args, **kwargs):
+
+        # What does this do? Seems to check if args are passed to decorator?
         is_bound_method = hasattr(args[0], wrapper.__name__) if args else False
         if is_bound_method:
             count = 1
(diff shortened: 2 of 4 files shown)


pymc4/utils.py
```diff
@@ -58,6 +58,8 @@ def merge_dicts(*dicts: dict, **kwargs: dict):
 def biwrap(wrapper):
     @functools.wraps(wrapper)
     def enhanced(*args, **kwargs):
+
+        # What does this do? Seems to check if args are passed to decorator?
```
canyon289: @ferrine 
For this biwrap utils, I get that biwrap  decorates `pm.coroutine.model` but what does enhance do specifically?

canyon289:
Questions from me might start slow at first but I hope to become more self sufficient as I gain more knowledge

ferrine:
You may check the docs here
https://github.com/ferrine/biwrap

pymc4/utils.py
```diff
@@ -55,18 +56,34 @@ def merge_dicts(*dicts: dict, **kwargs: dict):
...
+    Notes
+    -----
+    Currently this is only used to wrap pm.Model to capture model runtime flags such as
+    keep_auxiliary and keep_return. See pm.Model for all possible keyword parameters
+
+    """
+
     @functools.wraps(wrapper)
-    def enhanced(*args, **kwargs):
+    def enhanced(*args, **kwargs) -> Callable:
```
canyon289: @rpgoldman You've convinced me to start using type hints :). Did I do this right?

rpgoldman: That looks right. I believe that Callable can be specified further with arguments, though, if you know what arguments the returned function will take and/or you know what type it returns.
I'll have a peek now.

canyon289: Thanks for the quick answer, and adding type for function sounds fancy. I'll google around ot find examples.

Thanks again

pymc4/coroutine_model.py
```diff
@@ -61,21 +61,23 @@ class ModelTemplate:
...
     keep_auxiliary : bool
-        generative process may require some auxiliary variables to be created, but they are probably will not be used
-        anywhere else. In that case it is useful to tell PyMC4 engine that we can get rid of auxiliary variables
-        as long as they are not needed any more.
+        The generative process (``template``) may require some auxiliary variables to be created,
+        but they probably will not be used anywhere else. In that case it is useful to tell the
+        PyMC4 engine that it can get rid of them.
     keep_return : bool
-        the return value of the model will be recorded
+        The return value of the model will be recorded.
```
rpgoldman: Can you say where/how the return value will be recorded?

Also, maybe clarify what is meant by "return value", since this is a class not a function, and we have said that we use `yield`, which makes a stream of values instead of a return value.

pymc4/coroutine_model.py
```diff
@@ -91,9 +93,9 @@ def __call__(
 
         Parameters
         ----------
-        name : str
-            The desired name for the model, by default, it is inferred from the model declaration context,
-            but can be used just once
+        name : str or ``_no_name_provided``
```
rpgoldman: Could you add an example of the use of the `args` parameter?

canyon289: Would you mind if I add issues to capture your two questions and defer these an issue and then another PR? I personally don't know but am working through understanding PyMC4 by asking Maxim a bunch of questions?

rpgoldman: I would not mind at all, that sounds like a great idea.

pymc4/utils.py
rpgoldman: Are you missing an import for `Callable`?

canyon289: The import is here? 
https://github.com/pymc-devs/pymc4/pull/148/files#diff-03cbb4ef77afc157d399cf351453caf1R3

requirements-dev.txt
```diff
@@ -5,3 +5,4 @@ pytest-cov
 pytest-html
 pylint
 black
+mock
```
ferrine: Newline

canyon289:
> You can make it autouse with yield

I know, left it not autouse because there were only two and I was debating adding tests for model in a class with a different fixture at a future point

canyon289:
@ferrine This one good? I'll make issues for @rpgoldman's comments to keep track

ferrine:
Hi, small concerns about 
https://github.com/pymc-devs/pymc4/blob/master/tests/test_executor.py#L72

Better change to 
```
@model(method=True)
```
and 
https://github.com/pymc-devs/pymc4/blob/master/tests/test_executor.py#L82
in that case and not passing first argument which should be `self`

canyon289:
> Hi, small concerns about
> https://github.com/pymc-devs/pymc4/blob/master/tests/test_executor.py#L72
> 
> Better change to
> 
> ```
> @model(method=True)
> ```
> 
> and
> https://github.com/pymc-devs/pymc4/blob/master/tests/test_executor.py#L82
> in that case and not passing first argument which should be `self`

Noted. I can make those changes in another PR because theyre on master. Is this PR good to go?

tests/test_utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def _partial(*args, **kwargs):
+        raise Exception("Mocked functools partial")
+
+    _functools.partial = _partial
+
+    monkeypatch.setattr(pm.utils, "functools", _functools)
+
+
+def test_biwrao_and_mocked_functools_raises_exception_with_called_decorator(
```
twiecki: biwrao

twiecki:
Thanks @canyon289!

PR #149, opened 2019-10-05
Add Model as defined on a class

canyon289:
Anyone know what the if method code block does is intended for below? None of the tests hit that entire code block

### DIFF
--- a/pymc4/coroutine_model.py
+++ b/pymc4/coroutine_model.py
@@ -15,6 +15,7 @@ _no_name_provided = object()
 def model(genfn, *, name=_no_name_provided, keep_auxiliary=True, keep_return=True, method=False):
     """Flexibly wrap a generator function into a Model template."""
     if method:
+        # What is this block for?
         template = ModelTemplate(
             genfn, name=name, keep_auxiliary=keep_auxiliary, keep_return=keep_return
         )
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -68,13 +68,13 @@ def transformed_model_with_observed():
 
 @pytest.fixture("module")
 def class_model():
-    class PyMC4ClassModel:
+    class ClassModel:
         @pm.model
         def class_model_method(self):
             norm = yield pm.Normal("n", 0, 1)
             return norm
 
-    return PyMC4ClassModel()
+    return ClassModel()
 
 
 def test_class_model(class_model):
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -66,6 +66,27 @@ def transformed_model_with_observed():
     return transformed_model_with_observed
 
 
+@pytest.fixture("module")
+def class_model():
+
+    class PyMC4ClassModel:
+
+        @pm.model
+        def class_model_method(self):
+            norm = yield pm.Normal("n", 0, 1)
+            return norm
+
+    return PyMC4ClassModel()
+
+
+def test_class_model(class_model):
+    """Test that model can be defined as method in an object definition"""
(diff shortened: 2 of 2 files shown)


pymc4/coroutine_model.py
```diff
@@ -15,6 +15,7 @@
 def model(genfn, *, name=_no_name_provided, keep_auxiliary=True, keep_return=True, method=False):
     """Flexibly wrap a generator function into a Model template."""
     if method:
+        # What is this block for?
```
canyon289: This one code block :)

ferrine: In case first argument is self

ferrine: And if you do it another way, the defined method is not recognised as function, it is recognised as an object and self is not passed

ferrine: And if you do it another way, the defined method is not recognised as function, it is recognised as an object and self is not passed

ferrine: And if you do it another way, the defined method is not recognised as function, it is recognised as an object and self is not passed

canyon289: Thanks

tests/test_executor.py
```diff
@@ -66,6 +66,25 @@ def transformed_model_with_observed():
...
+    class PyMC4ClassModel:
+        @pm.model
+        def class_model_method(self):
+            norm = yield pm.Normal("n", 0, 1)
+            return norm
+
+    return PyMC4ClassModel()
+
+
+def test_class_model(class_model):
```
canyon289: Is this what it's supposed to look like? It works but its pretty ugly imo. Am I missing the more "pythonic" version?

canyon289:
Thanks @ferrine Ready for review and merge then

junpenglao:
I might have missed the discussion - what are we trying to do here? Is it to add support for any callable? to pm.sample (e.g., passing a class that has a `__call__` method)?

canyon289:
@ferrine has written logic that allows for PyMC4 models to be written as methods on a class.

I'm trying to understand all of PyMC4 by writing tests for everything and asking @ferrine questions

junpenglao:
Ah I see. In that case, could we create an arbitrary class in the test and monkey patch one of its method as a pymc4 model? In this case we can reuse the pymc4 models that we already defined in the same test.

canyon289:
> Ah I see. In that case, could we create an arbitrary class in the test and monkey patch one of its method as a pymc4 model? In this case we can reuse the pymc4 models that we already defined in the same test.

Now I am confused. The class, and instantiated object is in a fixture so it can be reused, and since this is testing user facing API we shouldnt need monkeypatch, after all the user won't be monkeypatching.

junpenglao:
Monkey patching just a way to save a few lines of defining the model in method again. But you are right, probably a more meaningful test case to spell out everything explicitly.

canyon289:
Ah I see what you mean. I think its better to leave it explicit for now so there's an example of class based API for models.

If this looks good can I merge?

junpenglao:
Thanks!

ferrine:
The main usecase of defining model within a method is being able to interchange model parts by subclassing.

Here is a draft of how it might look like. `other_method` is easy to override, but copy-pasting the whole model may be error prone 
```python
class BaseModel:
    @model
    def define_my_model(self, ...):
        var = yield self.other_method(...)

    def other_method(self, ...):
        raise NotImplementedError

class ModelType1(BaseModel):
    def other_method(self, ...):
        return (yield pm.Normal(...))
```

PR #152, opened 2019-10-09
Make tensorflow the only backend

ColCarroll:
This removes the ability to swap out backends in favor of only using tensorflow. I think it makes the code a bit easier to follow, but I probably missed a few things. I think this should be reviewed for any major problems, but otherwise we fix small things in followups.

I left in the ability to unpack distributions, and moved the distribution initialization to just after that (previously this had been done via multiple inheritance). I also made distribution initialization a static method in case we want to get the tensorflow distributions external to the pymc3 model for inspection.

I implemented a few new distributions along the way, and uncommented their tests.


pymc4/distributions/continuous.py
```diff
@@ -1,23 +1,189 @@
...
+    Parameter mappings to TensorFlow Probability are as follows:
+
+    - sigma: scale
+    """
+
+    def __init__(self, name, sigma, **kwargs):
+        super().__init__(name, sigma=sigma, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions):
```
canyon289: What's the point of these status methods if all they do is forward to the tfp distributions? I especially ask because in the next PR I'm going to remove the argument name differences so then they'll be exactly the same?

ColCarroll: For now, it was so that tests keep passing without changing them. From a larger point of view, I like that `unpack_conditions`, `_init_distribution`, and `_init_transform` provide developer hooks into initializing a random variable (the `Distribution.__init__` is below to provide context).

I might suggest keeping this, since then we can push out custom distributions without getting them into tensorflow probability (`HalfStudentT`, for example).

```python
        self.conditions = self.unpack_conditions(**kwargs)
        self._distribution = self._init_distribution(self.conditions)
        self.plate = plate
        super().__init__(
            self.unpack_distribution, name=name, keep_return=True, keep_auxiliary=False
        )
        if name is None and observed is not None:
            raise ValueError(
                "Observed variables are not allowed for anonymous (with name=None) Distributions"
            )
        self.model_info.update(observed=observed)
        self.transform = self._init_transform(transform)
        if self.plate is not None:
            self._distribution = tfd.Sample(self._distribution, sample_shape=self.plate)
```

canyon289: Keeping test working I understand

But for future case even in the above example though couldn't this line be 

```
self._distribution = tfd.some_dist(self.conditions)
```
if all that's happening is were initializing a tfd distribution?

ColCarroll: yep! that's even reasonable as a default. this is a bit of over engineering, but I think it may come in handy: 
- it is nice to get a raw `tfp` object to debug what is going on with a model using `pm.Normal._init_distribution(whatever)`
- there might be some shape handling that has to go on
- `self.conditions` might have some extra arguments
- `tfd` might not implement the distribution

ColCarroll: Er, to elaborate: I feel strongly that it should stay for this PR. I'm actually ok with removing it after merging this - you're right that it gets rid of some indirection and makes the intention more clear. I'd prefer it stayed for the above reasons, but I *do* recognize it is a little over-engineered!

twiecki:
This looks great, I'll go ahead and merge so that things go out of sync, if there are follow-up fixes we want to do it's no big deal.

PR #155, opened 2019-10-18
MAINT: use TFP parameter names

eigenfoo:
~WIP; do not merge.~ Ready for review.

Looks like the general (if begrudging, depending on who you talk to) consensus is to adhere to TFP parameter naming.


pymc4/distributions/continuous.py
```diff
@@ -337,26 +314,22 @@ class ChiSquared(PositiveContinuousDistribution):
...
+        Degrees of freedom (df > 0).
 
     Developer Notes
     ----------------
-    Parameter mappings to TensorFlow Probability are as follows:
-
-    - nu: df
-
     The ChiSquared distribution name is copied over from PyMC3 for continuity. We map it to the
     Chi2 distribution in TensorFlow Probability.
```
eigenfoo: According to this, we're calling it `ChiSquared` instead of `Chi2` to maintain continuity. Since nothing else is maintaining continuity, I suggest we also call this `Chi2`. Thoughts?

pymc4/distributions/continuous.py
```diff
@@ -1270,32 +1149,26 @@ class Uniform(BoundedContinuousDistribution):
...
     def _init_distribution(conditions):
-        lower, upper = conditions["lower"], conditions["upper"]
-        return tfd.Uniform(low=lower, high=upper)
+        low, high = conditions["low"], conditions["high"]
+        return tfd.Uniform(low=low, high=high)
 
+    # FIXME should we rename this functions as well?
     def lower_limit(self):
-        return self.conditions["lower"]
+        return self.conditions["low"]
```
eigenfoo: Should this function be called `lower_limit` or just `lower`?

pymc4/distributions/continuous.py
eigenfoo: I've gone ahead and done this in https://github.com/pymc-devs/pymc4/pull/155/commits/520459ff6cfde96bd8a0ab0f263749c6144adef9

I can revert that if there are objections 😄

pymc4/distributions/continuous.py
junpenglao: We can keep this naming, otherwise `lower` and `upper` alone might be confusing as some Distribution might have them as kwargs.

pymc4/distributions/continuous.py
twiecki: Agreed.

pymc4/distributions/continuous.py
```diff
@@ -81,33 +80,27 @@ class Normal(ContinuousDistribution):
...
-    Developer Notes
-    ---------------
-    Parameter mappings to TensorFlow Probability are as follows:
-
-    - mu: loc
-    - sigma: scale
+            x = pm.Normal('x', loc=0, scale=10)
     """
 
-    def __init__(self, name, mu, sigma, **kwargs):
```
twiecki: Now that we're just forwarding can't we move all this up into the base class and remove it here?

eigenfoo: I think @ColCarroll was opposed, for what I think is a pretty good reason: tab-completion and docstring-finders in many IDEs might rely on having explicit `__init__`s.

If we just had `*args, **kwargs`, then it would be hard for new users to find out the required parameters by e.g. running `pm.distributions.Normal??` in IPython.

pymc4/distributions/continuous.py
ferrine: If we bet bound to tfp, we better use domain objects from there

eigenfoo:
Also, as @twiecki points out, I haven't ported the example notebooks (like radon) yet. I'll add to this PR in the next few days. Sorry for the delay!

pymc4/distributions/continuous.py
eigenfoo: I'm not really sure why these functions exist: e.g. `Normal` doesn't have equivalent functions that return the same distribution parameters. Maybe I'm missing something, but is it necessary to have these functions at all?

eigenfoo:
@twiecki just reran the radon and eight schools notebooks. The eight schools notebooks actually raises some errors when you try to visualize the trace, so I didn't run all the cells, but everything up to (and including) the sampling works fine.

twiecki:
Thanks @eigenfoo!

PR #156, opened 2019-10-18
MAINT: minor improvements

eigenfoo:
1. Adds environment directories that are generated from the `Makefile` to the `.gitignore`
2. Lowercase package title (consistent with [PyMC3](https://github.com/pymc-devs/pymc3/blob/master/setup.py#L7) and Python packaging conventions) and add classifiers and license to `setup.py`
3. Add `FUNDING.yml` to the `.github` directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,9 @@ docs/_build/
 .env
 .venv
 env/
+pymc4-env/
 venv/
+pymc4-venv/
 ENV/
 env.bak/
 venv.bak/
--- a/setup.py
+++ b/setup.py
@@ -11,11 +11,25 @@ REQUIREMENTS_DEV_FILE = PROJECT_ROOT / "requirements-dev.txt"
 README_FILE = PROJECT_ROOT / "README.md"
 VERSION_FILE = PROJECT_ROOT / "pymc4" / "__init__.py"
 
-NAME = "PyMC4"
+NAME = "pymc4"
 DESCRIPTION = "A Python probabilistic programming interface to TensorFlow, for Bayesian modelling and machine learning."
 AUTHOR = "PyMC Developers"
 AUTHOR_EMAIL = "pymc.devs@gmail.com"
 URL = ("https://github.com/pymc-devs/pymc4",)
+LICENSE = "Apache License, Version 2.0"
+
+CLASSIFIERS = [
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.5",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -11,11 +11,25 @@
 README_FILE = PROJECT_ROOT / "README.md"
 VERSION_FILE = PROJECT_ROOT / "pymc4" / "__init__.py"
 
-NAME = "PyMC4"
+NAME = "pymc4"
```
canyon289: just curious what differences does this make? Does it install as "PyMC4" with pip or vnenv in the config?

canyon289:
LGTM

setup.py
twiecki: I think pypi uses this

.gitignore
```diff
@@ -55,7 +55,9 @@ docs/_build/
 .env
 .venv
 env/
+pymc4-env/
 venv/
+pymc4-venv/
```
twiecki:
```suggestion
```

eigenfoo: Actually, this was intentional: the Makefile creates these directories: https://github.com/pymc-devs/pymc4/blob/master/Makefile#L28

I almost committed it while making this PR 😄

twiecki:
Yah but the line appeared twice.

eigenfoo:
https://github.com/pymc-devs/pymc4/pull/156/files#diff-a084b794bc0759e7a6b77810e01874f2R58-R60

One is `venv` (from Python venvs), and one is `env` (from conda envs).

twiecki:
Oh, sorry, misread that.

eigenfoo:
No worries! This PR looks like it's good to go then?

-------- Original Message --------

twiecki:
Yeah let's merge, I'm not convinced the tab completion argument is good
enough but we can revisit that later.

twiecki:
Did you adapt the examples like radon?

eigenfoo:
Lol, this is the minor improvements PR. You must mean the parameter renaming PR, where I mentioned tab completion?

I'll go ahead and merge this PR.

~ George

-------- Original Message --------

rpgoldman:
> Yeah let's merge, I'm not convinced the tab completion argument is good enough but we can revisit that later.

For the record, I'm with @eigenfoo on this one.  I think of functions with `(*args, **kwargs)` can be terribly difficult to learn as a user -- the environment really doesn't help you, and often the only way to figure out how to use something is to search an entire inheritance tree -- or even worse if there's delegation + inheritance.

It also supports bugs where you misspell something (e.g. `std` for `sd`), and the system quietly goes off and applies a default.  These can be quite hard to debug.

Use of `**kwargs` is often unavoidable, but I favor avoiding them as much as possible, and providing regular keyword arguments whenever possible.

PR #158, opened 2019-10-18
Add log_prob_elemwise feature

### LINKED ISSUE #157: [Feature request] add a log_prob_elemwise api
After some internal discussion, we agreed that it would be good to be able to get a model's element-wise `log_prob` evaluation. This feature would help with:

- Debugging shape problems
- Finding pathological point values in the observed data passed onto random variables

lucianopaz:
This closes #157.

### Core idea

I simply copied the ordinary `SamplingExecutor.collect_log_prob` to return a dictionary of `{"rv_name": elem_log_prob_value}` instead of reducing the `dist.log_prob(...)` with `tf.reduce_sum`.

Using this we can use `pm.evaluate_model(model())` to get a `SamplingState` which can then be used to `collect_log_prob_elemwise`.

### Ease of use

I think that it should be easy for an end-user to be able to get the output of `collect_log_prob_elemwise` without having to jump through so many hoops (evaluating the model, getting a `SamplingState`, finally calling the `SamplingState.collect_log_prob_elemwise`) so I added `model_log_prob_elemwise` which does all of that behind the scenes.

### Discussion

- `model_log_prob_elemwise` isn't really needed. It's just there to help end-users. Should it be moved to another part of pymc4?
- Is the interface I wrote for `model_log_prob_elemwise` OK?
- When we decide on `model_log_prob_elemwise`'s fate, I'll write it's docstring
- It would be great to test this with some multivariate distribution, where the `event_shape` isn't `()`.


kyleabeauchamp:
Thanks, will play with this PR soon.

kyleabeauchamp:
I'm having reasonable success with this approach.  For various reasons, I ended up needing to do something like the following:

```
    for i in range(n_steps):
        tf_trace = pm4.inference.sampling.sample(model_instantiated, ...)
        lp = pm4.model_log_prob_elemwise(model_instantiated)
        # Accumulate arrays etc
```

Eventually I'll try to add this as a deterministic inside the model, once that feature stabilizes.

kyleabeauchamp:
Actually, I guess if I pass `values=trace[0]` to the `model_log_prob_elemwise` function, is it going to run the calculation for each step in the MCMC trace?  I guess that avoids the above for loop that I mentioned, if I'm current in my interpretation of what it's intending to do.

lucianopaz:
Adding as a deterministic might be a problem. I'll ping @ferrine to hear his input, but within the model function's flow, the distribution instance is never available. The model function only sees the samples that come out of the distribution, i.e. the values.
To be able to see the distribution, we would need the model to somehow have a reference to the `SamplingState`.

kyleabeauchamp:
I think I'm fine without deterministic at this point.

lucianopaz:
About your second point, I think that yes, you should be able to pass `values=trace[0]` to `model_log_prob_elemwise`. I haven't tried it out yet, though.

lucianopaz:
I rebased this and **removed** the high level `model_log_prob_elemenwise` API. In the end, this didn't look important to me, as we can wrap the distribution's `log_prob` in a `Deterministic` in the following way:

```python
@pm.model
def model():
    a = yield pm.Normal("a", 0, 1)
    b = yield pm.HalfNormal("b", 1)
    c_dist = pm.Normal("c", loc=a, scale=b, observed=np.random.randn(100))
    c = yield c_dist
    log_prob = yield pm.Deterministic("log_prob", c_dist._distribution.log_prob(c))
```

The key to this method is that we don't yield from the `pm.Distribution` immediately, so **we have access to the underlying `tfp.Distribution`**. This would also allow us to use the batch and event shape manipulations that `tfp` provides, such as `Independent`, `BatchReshape` and others.

On the other hand, I added a function to get an iterable over all the distribution and potential's unreduced `log_prob` values. This could prove useful if we want to add some way to allow users to manually handle batching to improve runtimes.

pymc4/flow/executor.py
```diff
@@ -143,12 +143,17 @@ def __init__(
...
+        return itertools.chain(
             (dist.log_prob(self.all_values[name]) for name, dist in self.distributions.items()),
             (p.value for p in self.potentials),
         )
-        return sum(map(tf.reduce_sum, all_terms))
+
+    def collect_log_prob(self):
+        return sum(map(tf.reduce_sum, self.collect_log_prob_elemwise()))
+
+    def collect_unreduced_log_prob(self):
```
ColCarroll: this looks like a `ValueError` waiting to happen, unless all the elements are the same shape, right?

tests/test_executor.py
```diff
@@ -509,6 +561,21 @@ def model():
...
+def test_log_prob_elemwise(fixture_model_with_plates):
+    model, expected_rv_shapes = fixture_model_with_plates
+    _, state = pm.evaluate_model(model())
+    log_prob_elemwise = dict(
+        zip(state.distributions, state.collect_log_prob_elemwise())
+    )  # This will discard potentials in log_prob_elemwise
+    log_prob = state.collect_log_prob()
+    assert len(log_prob_elemwise) == len(expected_rv_shapes)
+    assert all(rv in log_prob_elemwise for rv in expected_rv_shapes)
+    print({k: v.shape for k, v in log_prob_elemwise.items()})
```
ColCarroll: can remove prints

pymc4/flow/executor.py
lucianopaz: Yes

We are doing automatic batching in all three big sampling functions: `sample_prior_predictive`, `sample` and `sample_posterior_predictive`. To be able to get the auto batching to work in some cases brings an over-head of extra `evaluate_model*` calls. Furthermore, the auto batching will have the overhead of always creating `SamplingState`s and other intermediate entities for every batch element. So in the end, I think that if someone wants to get their hands dirty to manually ensure that their model is well vectorized and batched, then they should have the means to do so.

I agree that this function is dangerous. It could be a `ValueError` waiting to happen, or a silently well broadcasting thing that leads to mysterious errors down the way, but it provides the necessary interface for a user to manually vectorize and batch their model.

tests/test_executor.py
lucianopaz: Good catch

pymc4/flow/executor.py
lucianopaz: It's very hard to get the model to be well vectorized. If there are reductions, or tensor products, or even simple sums, the `batch_shape`s can lead to horribly wrong `log_prob` tensors. The automatic batching is super user friendly and awesome, but someone may ask for a more low level control of what goes on

ColCarroll: i'm fine leaving this -- was just wondering if I was right or wrong about it being dangerous :D

PR #159, opened 2019-10-27
BLD: implement auto-naming

### LINKED ISSUE #122: auto_name doesn't work if a line contains more than just the RV specifciation
Model:

```python
@pm.model(auto_name=True)
def model():
    ....
    # The offending line is this one.
    nu = pm.Exponential(lam=1 / 29.0) + 1
    ...
```

Error message:

```
________________________________________________________________________ ERROR collecting pymc4/tests/test_hierarchical_comparison.py ________________________________________________________________________
pymc4/tests/test_hierarchical_comparison.py:41: in <module>
    model = model.configure()
pymc4/_model.py:64: in configure
    model._evaluate()
pymc4/_model.py:80: in _evaluate
    self._template._func(*args, **kwargs)
pymc4/tests/test_hierarchical_comparison.py:33: in model
    nu = pm.Exponential(lam=1 / 29.0) + 1
pymc4/random_variables/random_variable.py:134: in __init__
    raise ValueError("No name was set. Supply one via the name kwarg.")
E   ValueError: No name was set. Supply one via the name kwarg.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================================================== 52 deselected, 1 error in 1.79 seconds ===================================================================================
```

eigenfoo:
WIP. Closes #122.

Some light refactoring of the existing `ast_compiler.py` (mainly to remove unnecessary custom errors and exceptions).

The main contribution here is the `parse_random_variable_names` function in `ast_compiler.py`, and relevant tests.

```python
import pymc4 as pm

def example():
    x = yield pm.Normal(loc=0, scale=1)

names = parse_random_variable_names(example)
assert names == ["x"]
```

I still need some help navigating [the `coroutine_model.py` module](https://github.com/pymc-devs/pymc4/blob/master/pymc4/coroutine_model.py#L15). What should I do to make the `auto_name` flag that we want?


pymc4/ast_compiler.py
```diff
@@ -114,26 +103,68 @@ def parse_snippet(source, filename, mode, flags, firstlineno, privateprefix_igno
...
+
+        # We expect the yielded expression to be a function call. If it is not,
+        # raise an exception.
+        if not isinstance(yielded, ast.Call):
+            msg = "Unable to auto-name: expected all yielded expressions to be function calls."
+            raise RuntimeError(msg)
+
+        # We expect there to be only one target. If there are more, raise an
+        # exception.
+        if len(names) > 1:
```
twiecki: What's cool is that this acts as a pymc4 syntax checker.

pymc4/ast_compiler.py
```diff
@@ -114,26 +103,68 @@ def parse_snippet(source, filename, mode, flags, firstlineno, privateprefix_igno
...
-                return tree_node
-            else:
-                tree_node.value.keywords.append(ast.keyword("name", ast.Str(rv_name)))
-        except AttributeError:
-            pass
-
-        return tree_node
+class AutoNameVisitor(ast.NodeVisitor):
+    def __init__(self):
+        self.random_variable_names = []
```
twiecki: Don't you have to call `super()`?

twiecki: In fact, it might be better to not overload `__init__` and just return `random_variable_names` from visit.

eigenfoo: So you mean, overload the `visit` function? How would we pass the names from the `visit_Assign` calls to the overall `visit` call?

eigenfoo:
@twiecki I'm still a bit confused: where in [1] should I add the `auto_name` flag? And should I delete [2] now that we have a way of parsing out names from the model function?

[1] https://github.com/pymc-devs/pymc4/blob/master/pymc4/coroutine_model.py#L15
[2] https://github.com/pymc-devs/pymc4/blob/master/pymc4/coroutine_model.py#L35

pymc4/ast_compiler.py
eigenfoo: I've added a `super(...).__init__()` call for now.

eigenfoo:
I'm also a bit confused about what you mentioned in #122 

> Obviously this doesn't allow for models that change their RV order but that's a fine limitation. Some care needs to be taken care with model nesting but I think if we have that logic in the `@model` decorator we should be able to get a clean implementation.

What do you mean by "model nesting"? And is there anything I can do to detect "models that change their RV order", and raise an appropriately helpful error message?

pymc4/ast_compiler.py
twiecki: No, keep it like it is but return something in `visit_Assign`, if that's possible. That you wouldn't need the state and could delete `__init__`.

twiecki:
With model nesting I mean submodels, i.e. calling a model from within another model, like in the last section of the radon model to create the hyperpriors.

I don't think we can do anything to detect it if the number of RVs stays the same. But I don't think anyone is doing that really and would expect it to just work, also the sampler would choke if the dimensionality of the model changed.

pymc4/ast_compiler.py
eigenfoo: Hm, I don't think that's how the `visit_Assign` method works: it gets called once per assignment statement (or in our case, once per `yield`), not once per model parsing, so the only thing it could do is return one name at a time, which isn't what we want. I think that overriding `__init__` is the closest thing we have.

twiecki:
Regarding the implementation, we might want to tackle that in a separate PR, but I think it would probably need to go somewhere here: https://github.com/pymc-devs/pymc4/blob/master/pymc4/flow/executor.py#L388. First we parse the names and store them as a list, then as we iterate over the model-generator where we intercept the distributions we set the `.name` attribute on them with the name provided by the list. Unfortunately that code is already quite convoluted and this will add to that logic. Maybe there is a better way but that would be the obvious place to start.

pymc4/ast_compiler.py
twiecki: You're right.

twiecki:
> I don't think we can do anything to detect it if the number of RVs stays the same.

Technically we could store the type of RV being referenced in the AST and make sure they match when we assign names, but I don't think it's worth it.

PR #160, opened 2019-10-30
BLD: add arviz integration

### LINKED ISSUE #5: Add arviz integration
Add `traceplot()` and `plot_posterior()` functions.

eigenfoo:
Closes #5.

This is probably getting ahead of ourselves, but this is a simple callthrough to functions from ArviZ.


eigenfoo:
Hm, looks like black is failing on some files that this PR doesn't modify. Any ideas why?

aloctavodia:
I think we should use ArviZ directly.

twiecki:
Yeah, I agree, hiding this will cause unnecessary redirection and create confusion further down the line.

PR #162, opened 2019-11-21
Use codecov for code coverage

lucianopaz:
Codecov was a nice replacement for coveralls in pymc3. This PR changes the currently used azure-pipelines coverage publisher and reporter tasks to codecov.io instead. We'll get the same nice coverage graphs for PRs and reports as in pymc3.

I also added a python3.8 test environment (as it's basically free because we haven't hit the limit of parallel jobs in azure).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,7 +9,7 @@ strategy:
       python.version: "3.6"
     Python37:
       python.version: "3.7"
-    Python37:
+    Python38:
       python.version: "3.8"
 
 steps:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,6 +9,8 @@ strategy:
       python.version: "3.6"
     Python37:
       python.version: "3.7"
+    Python37:
+      python.version: "3.8"
 
 steps:
   - task: UsePythonVersion@0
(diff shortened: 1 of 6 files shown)


lucianopaz:
Tensorflow and TFP don't seem to support python3.8 yet... I'll just drop the python3.8 test env entirely until they package TF and TFP for py38

azure-pipelines.yml
```diff
@@ -44,14 +44,7 @@ steps:
       python -m pytest -xv --cov pymc4 --junitxml=junit/test-results.xml --cov-report xml --cov-report term --cov-report html .
     displayName: "pytest"
 
-  - task: PublishTestResults@2
+  - script: |
+      bash <(curl -s https://codecov.io/bash)
```
ColCarroll: neat! does this just run a bash script that will check for $CODECOV_TOKEN?

lucianopaz: Yup. At least it doesn't seem to be required on [azure](https://docs.codecov.io/docs/about-the-codecov-bash-uploader)

lucianopaz:
I made an almost empty commit to get the first coverage report uploaded properly to codecov

PR #161, opened 2019-11-18
Implement deterministics

lucianopaz:
This PR builds on #147 but goes beyond it to implement deterministics that are traced while sampling. It does so by building a callback function to compute the deterministic after each kernel step.

I added a few tests, but XLA sampling does not work. So this PR can be a starting point for deterministics in pymc4 to build upon.

PR #164, opened 2019-11-22
[WIP] Implement high level forward sampling API

### LINKED ISSUE #163: Add high level sample_prior_predictive and sample_posterior_predictive
We've got a high level `sample` function skeleton in place, but it would be nice to also include a high level API like pymc3's `sample_prior_predictive` and `sample_posterior_predictive`.

lucianopaz:
This PR closes #163.
It is based on top of #161.

It aims to implement two high level API functions:

* [x] `sample_prior_predictive`
* [x] `sample_posterior_predictive`

---
## Sample prior predictive

`sample_prior_predictive` will take a `Model` and the desired `sample_shape`, and returns draws for every:
* untransformed distribution
* observed distribution
* deterministic
present in the model's control flow. It will rely on `pymc4.flow.evaluate_model` to draw samples from the `Model` instance.

While the model is evaluated, the distributions' observed values will be ignored by default. This behavior can be overiden, and the supplied observed values will be used without any checks to their validity.

---
## Sample posterior predictive

`sample_posterior_predictive` will take a `Model` and a `Trace` of sampled values. There should be two supported scenarios:

1. The `Trace` could be the output from `pm.sample`, and thus will include values for untransformed distributions, deterministics and **ONLY** the values on the entire real line for the auto-transformed distributions (i.e. values for `__log_dist` will be stored but not `dist`).
2. The `Trace` could be part of the output from a `sample_prior_predictive` call, and will not include any transformed distributions.

The other requirement is that the `observed` values of a distribution must be taken into account to determine the distribution's shape. For example, in a glm:

```python
@pm.model
X = np.empty((n_observations, n_features))
y = np.empty((n_observations,))
def glm():
    beta = yield pm.Normal("beta", 0, 1, plate=n_features)
    obs = yield pm.Normal(
        "obs",
        tf.tensordot(tf.convert_to_tensor(X), beta, axes=[[1], [0]]),
        0,
        observed=y
    )
```

The `obs` distribution shape will be inferred from the parameters to be `(n_observations)`, which matches with `y.shape`.

In a simpler scenario:

```python
@pm.model
y = np.empty((n_observations,))
def glm():
    beta = yield pm.Normal("beta", 0, 1)
    obs = yield pm.Normal("obs", beta, 0, observed=y)
```

The distribution shape of the `obs`'s variable will be `()`. The `observed=y` should fully determine `obs`'s shape in `sample_posterior_predictive`.

To do this, we will use a combination of a new executor (like `pymc4.flow.transformed_executor.TransformedSamplingExecutor`) and either overwriting the observed `_distribution` attribute or using `tf.Independent` to change the `batch_shape`.

---
## Progress

* [x] `sample_prior_predictive`
* [x] `sample_prior_predictive` tests
* [x] `sample_prior_predictive` docs
* [x] `sample_posterior_predictive`
* [x] posterior predictive executor
* [x] posterior predictive trace handling
* [x] `sample_prior_predictive`, executor and trace handling tests
* [x] `sample_prior_predictive`, executor and trace handling docs


lucianopaz:
It looks like trace handling wont be an issue. The `TransformedSamplingExecutor` already handles the execution when we feed in samples from the prior or from the traced posterior. It does so with the `values` kwarg, like in the following examples:

```python
prior = pm.sample_prior_predictive(model(), 10)
_, state = pm.evaluate_model_transformed(model(), values=prior)
```

```python
trace, stats = pm.inference.sampling.sample(model(), 10)
_, state = pm.evaluate_model_transformed(model(), values=trace)
```

The evaluated `state` has the correct auto transformed values for both cases.

lucianopaz:
I think that this is ready for review. The patch's code coverage is high (97%) and the only two lines missed are not vital.

lucianopaz:
On second thought, I can make a slight change to `assert_observations_compatible_with_distribution_shape` to get the patch coverage up to 100%. I'll get it pushed tomorrow

lucianopaz:
Thanks @colcarroll! I'll merge then so people can start to use the API.

PR #165, opened 2019-11-23
[WIP] Attempt to fix azure triggers for branches

canyon289:


azure-pipelines.yml
```diff
@@ -1,5 +1,12 @@
 trigger:
-  - master
+  branches:
+    include:
+      - master
+      - releases/*
+  pr:
```
lucianopaz: From the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml) it looks like `trigger` and `pr` should both be at the root level of the yaml.

```suggestion
pr:
  branches:
    include:
      - master
      - releases/*
```

Out of curiosity, why don't we want to run the pipeline for PRs that don't target master or releases?

canyon289: 🤦 I mean to open this on ArviZ which is having double builds. Closing here. Apologies on my mistake

PR #172, opened 2019-12-04
Remove AST compiler and autonaming modules

eigenfoo:
Remove the AST compiler module, since we're no longer trying to parse the AST to hide the `yield` keyword.


ColCarroll:
:fire: :computer: :fire:

twiecki:
But we can still do autonaming, no?

PR #173, opened 2019-12-05
Add dtype to distribution

ColCarroll:
This is for developer comfort -- there are still a fair number of bugs from type mis-matches, and it is nice to be able to do `var.dtype` instead of `var._distribution.dtype`.


lucianopaz:
Wouldn't a `@property` setter and getter be better here? Maybe someone might change part of the `_distribution`  attribute

ColCarroll:
Yes! Good point.

ColCarroll:
Oh, I only added a getter for now, since `tfp` also does not supply a setter (and I don't want one yet :) )

LMK if you think we should add it anyways.

lucianopaz:
Thanks @colcarroll! Looks great!

Issue #170, opened 2019-12-03
Error is not captured for models with no free RVs


rrkarim:
There is no exception handling for models with only observed variables. Maybe one can add `ndim` method to the state object to make all the checks in `sample` method but there is some ambiguity with shape support #91? How to better fix the issue?

lucianopaz:
Would you like `inference.sampling.sample` to raise an error under these circumstances? It should be easy to do in `build_model_logp*`. Would you like to try a PR?

rrkarim:
Yeah, sure, I thought that it might be clear if all the checking is done in `sample` method..

PR #175, opened 2019-12-05
Fix exception handling for no free RVs

rrkarim:
#170
plus make the logic bit easier.

### DIFF
--- a/pymc4/inference/sampling.py
+++ b/pymc4/inference/sampling.py
@@ -170,7 +170,7 @@ def build_logp_and_deterministic_functions(
     state, deterministic_names = initialize_sampling_state(model, observed=observed, state=state)
 
     if not state.all_unobserved_values:
-        raise ValueError("The `state` doesn not contain any `unobserved` values")
+        raise ValueError("The `state` does not contain any `unobserved` values")
 
     observed = state.observed_values
     unobserved_keys, unobserved_values = zip(*state.all_unobserved_values.items())
--- a/pymc4/inference/sampling.py
+++ b/pymc4/inference/sampling.py
@@ -166,12 +166,11 @@ def build_logp_and_deterministic_functions(
         )
     if state is not None and observed is not None:
         raise ValueError("Can't use both `state` and `observed` arguments")
-    if state is None:
-        state, deterministic_names = initialize_state(model, observed=observed)
-    else:
-        _, st = flow.evaluate_model_transformed(model, state=state)
-        deterministic_names = list(st.deterministics)
-        state = state.as_sampling_state()
+
+    state, deterministic_names = initialize_state(model, observed=observed, state=state)
+
+    if not state.all_unobserved_values:
+        raise ValueError("The `state` doesn not contain any `unobserved` values")
 
     observed = state.observed_values
     unobserved_keys, unobserved_values = zip(*state.all_unobserved_values.items())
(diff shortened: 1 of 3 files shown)


pymc4/inference/sampling.py
```diff
@@ -166,12 +166,11 @@ def build_logp_and_deterministic_functions(
...
-        state, deterministic_names = initialize_state(model, observed=observed)
-    else:
-        _, st = flow.evaluate_model_transformed(model, state=state)
-        deterministic_names = list(st.deterministics)
-        state = state.as_sampling_state()
+
+    state, deterministic_names = initialize_sampling_state(model, observed=observed, state=state)
+
+    if not state.all_unobserved_values:
+        raise ValueError("The `state` does not contain any `unobserved` values")
```
ColCarroll: Can this message be a bit more informative? Maybe:

```python
raise ValueError(f"Can not calculate a lot probability: the model {model.name or ""}  has no unobserved values.")
```

tests/test_sampling.py
```diff
@@ -172,3 +182,12 @@ def test_sampling_with_deterministics_in_nested_models(
...
         np.testing.assert_allclose(trace[deterministic], op(*[trace[i] for i in inputs]), rtol=1e-6)
+
+
+def test_sampling_with_no_free_rvs(simple_model_no_free_rvs):
+    model = simple_model_no_free_rvs()
+    with pytest.raises(ValueError) as error:
+        trace, stats = pm.inference.sampling.sample(
+            model=model, num_samples=1, num_chains=1, burn_in=1
+        )
+    error.match("The `state` does not contain any `unobserved` values")
```
ColCarroll: If you're updating the message above, I am ok not validating the error message, since that can be brittle.

pymc4/inference/sampling.py
ColCarroll: Ugh. "Can not calculate a *log* probability"

rrkarim: Yeah sure, then there is not point to include the additional test case.

pymc4/inference/sampling.py
```diff
@@ -166,12 +166,13 @@ def build_logp_and_deterministic_functions(
         )
     if state is not None and observed is not None:
```
rrkarim: This one can also be removed since there is a check in evaluate_model but the exception there is not very intuitive for me.

pymc4/inference/sampling.py
ColCarroll: I thought `with pytest.raises(ValueError)` was a good test case, and confirms the issue was fixed

rrkarim: ok, anyway codecov doesn't let me to do it.

pymc4/inference/utils.py
```diff
@@ -4,8 +4,8 @@
 from .. import Model, flow
 
 
-def initialize_state(
-    model: Model, observed: Optional[dict] = None
+def initialize_sampling_state(
+    model: Model, observed: Optional[dict] = None, state=None
```
lucianopaz: You forgot to add `state` into the evaluate model call. It should also be noted in the docstring

ColCarroll:
Thanks @rrkarim!

lucianopaz:
Oh, my review arrived too late. Well, `initialize_sampling_state` can be fixed in another pr

rrkarim:
@lucianopaz my bad, fixing.

PR #178, opened 2019-12-06
Half Student T

ColCarroll:
This adds the half student t distribution in preparation for adding a port of [the rugby case study](https://docs.pymc.io/notebooks/rugby_analytics.html) from pymc3.

I also changed the requirements to use nightly builds of tensorflow and tfp again, since that actually makes some tests turn into "unexpected pass".

Note the coding style is a little strange: I plan to make a PR with the half student t class upstream to tensorflow probability shortly, so I wrote it with that style in mind.


ColCarroll:
This is helpful for checking out transforms, so merging!

junpenglao:
+1 for it to go upstream.

PR #180, opened 2019-12-07
Untransform variables in inference

ColCarroll:
So this effectively adds the transformed variables back to the stack of deterministics. It caused an error in `sample_posterior_predictive`, but if I just... ignore the error, everything seems to work fine. In particular tests pass.

I am going to look at this more later, just wanted to show some WIP.

Some unrelated whitespace changes and debugging got in as well that I'll remove.

### DIFF
--- a/pymc4/flow/transformed_executor.py
+++ b/pymc4/flow/transformed_executor.py
@@ -4,10 +4,19 @@ Specifically, we wish to transform distributions whose support is bounded on
 one or both sides to distributions that are supported for all real numbers.
 """
 import functools
+
+
+from typing import Mapping, Any
 from pymc4 import scopes, distributions
 from pymc4.distributions import distribution
 from pymc4.distributions.transforms import JacobianPreference
-from pymc4.flow.executor import SamplingExecutor, EvaluationError, observed_value_in_evaluation
+from pymc4.flow.executor import (
+    SamplingExecutor,
+    EvaluationError,
+    observed_value_in_evaluation,
+    ModelType,
+    SamplingState,
+)
 
 
 class TransformedSamplingExecutor(SamplingExecutor):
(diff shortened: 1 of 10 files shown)


pymc4/flow/transformed_executor.py
```diff
@@ -55,17 +55,17 @@ def modify_distribution(self, dist, model_info, state):
...
-                transformed_scoped_name in state.transformed_values
-                and scoped_name in state.untransformed_values
-            ):
-                raise EvaluationError(
-                    "Found both transformed and untransformed variables in the state: "
-                    "'{} and '{}', but need exactly one".format(
-                        scoped_name, transformed_scoped_name
-                    )
-                )
+            # # 0. do not allow ambiguity in state, make sure only one value is provided to compute logp
```
lucianopaz: I think that this shouldn't be commented out. Given that this PR will make our `trace`s always have the untransformed variables, we could make `PosteriorPredictiveSamplingExecutor` inherit from `SamplingExecutor` instead of `TransformedSamplingExecutor`.

ericmjl:
@ColCarroll @lucianopaz anything stopping this PR?

ColCarroll:
i think luciano was suggesting that i make a separate subclass for the posterior predictive executor, which keeps the exception in place. 

i think this is very close (i also wanted to make sure it was tested), and it is now next on my list to get over the finish line, but life is busy until january! i won't be offended if someone wants to push to this and get it merged.

ericmjl:
@ColCarroll understood. I'm also going to be out on vacation for a while. If this isn't picked up by the time I'm back, I'll ping you to see what I can help with.

lucianopaz:
Yeah, I'll also be away the next week. Until January ain't nobody got time for this!

As a side note, @colcarroll, there already is a special executor for posterior predictive sampling in the flow folder. The thing is that it inherits from `TransformedSamplingExecutor` instead of `SamplingExecutor`. I chose to do that to be able to do posterior predictive sampling from the trace returned by `sample` which had the transformed values. If we change `sample` we can also make the posterior predictive executor inherit from the base `SamplingExecutor`.

ColCarroll:
This refactors a bunch, but also removes the bit that actually transforms/untransforms variables. I'll add that back in soon.

pymc4/flow/transformed_executor.py
```diff
@@ -17,124 +26,112 @@ def validate_state(self, state):
...
 
-        if dist.transform is None or dist.model_info.get(  # do nothing else if no transform is set
-            "autotransformed", False
-        ):  # already autotransformed, do nothing else
-            return dist
-        transform = dist.transform
-        transformed_scoped_name = scopes.variable_name(
-            # double underscore stands for transform
-            "__{}_{}".format(transform.name, dist.name)
+        return transform_dist_if_necessary(dist, state, allow_transformed_and_untransformed=True)
```
ColCarroll: here is where most of the change happened -- I factored this logic into 3 functions, with this one switch, and had both executors use it.

tests/test_8schools.py
```diff
@@ -18,9 +21,51 @@ def schools_pm4():
...
+    )
+    init_value = logp(
+        **{
+            "schools_pm4/eta": tf.zeros(8),
+            "schools_pm4/mu": tf.zeros(()),
+            "schools_pm4/__log_tau": tf.ones(()),
+        }
+    ).numpy()
+
+    assert_almost_equal(init_value, -42.876_114)
```
ColCarroll: note this new test makes sure that the log probability is correct (for at least one point!)

ColCarroll:
This is a funny failure -- broadcasting error on Python 3.6 but not 3.7. I can recreate it on my machine, and I see that the two models all have the same shapes...

ColCarroll:
Extra fun: on python 3.6, the failure is intermittent

pymc4/flow/transformed_executor.py
```diff
@@ -17,124 +26,112 @@ def validate_state(self, state):
...
-                transformed_scoped_name in state.transformed_values
-                and scoped_name in state.untransformed_values
-            ):
-                raise EvaluationError(
-                    "Found both transformed and untransformed variables in the state: "
-                    "'{} and '{}', but need exactly one".format(
-                        scoped_name, transformed_scoped_name
-                    )
+            raise EvaluationError(
+                EvaluationError.OBSERVED_VARIABLE_IS_NOT_SUPPRESSED_BUT_ADDITIONAL_TRANSFORMED_VALUE_PASSED.format(
```
junpenglao: wow the error message LOL

tests/test_8schools.py
```diff
@@ -18,9 +21,51 @@ def schools_pm4():
...
+    with pm.Model() as eight_schools:
+        eta = pm.Normal("eta", 0, 1, shape=J)
+        mu = pm.Normal("mu", 1, 10)
+        tau = pm.HalfNormal("tau", 2.0)
+
+        theta = mu + tau * eta
+
+        pm.Normal("obs", theta, sigma, observed=y)
+
+    print(eight_schools.logp({'eta': np.zeros(8), 'mu': 0, 'tau_log__': 1}).astype(np.float32))
```
junpenglao: tau_log__ --> __log_tau according to below

ColCarroll: `tau_log__` is from PyMC3! We spell it `__log_tau` now! :grin:

ColCarroll:
So, things were failing because of https://bugs.python.org/issue32792 .

I grabbed the current source-code and "backported" it. this is a heck of a hack, but the only downside I could see is if Python 3.x changes how `ChainMap.__iter__` is implemented in a major way, in which case our tests would fail anyways.

ColCarroll:
Merging so I can keep making changes. Feel free to request updates here (or revert, even!) and I can address them after.

PR #185, opened 2019-12-16
Added `sample` to pymc4 top-level namespace.

ericmjl:
This is a tiny PR that will allow us to do:

```python
import pymc4 as pm

# model definition here
model = ...

trace, diagnostics = pm.sample(model)
```


ericmjl:
@twiecki @ColCarroll and @ferrine this is ready for your review.

ferrine:
LGTM

Issue #169, opened 2019-12-03
Currently there are transformed values captured in trace object


rrkarim:
We don't see untransformed values (if distribution requires transformation) in trace object, since `initialize_state`  turns `state` into `sampling_state`. Maybe there is a need to capture `untransformed` values before `initialize_state` call and then inverting appropriate RVs to get more convenient trace. How one can implement it better?

rrkarim:
Also why not renaming `initialize_state` to `initialize_sampling_state` to remove redundant code in `build_logp_and_deterministic_functions` with all the state/observed checks?

twiecki:
Those sound like reasonable changes and requests, happy to review PRs!

rrkarim:
Working on it, guess there are some related issues to fix first.

ColCarroll:
@rrkarim are you making progress on this? i may try to tackle it today. I want to be able to just

```python
data = pm.sample(model())
```
and have `data` be the arviz object. This will require:

- Update import paths to get `sample` at the top level
- Wrap the return value of `sample` into the `arviz` object
- Add the untransformed values into the trace (probably as Deterministics? this looks at least a little tricky)

I think the first two will be easy, and I will tackle those first.

rrkarim:
Yes, the third one is, indeed, looks little tricky with my knowledge of the current design requirements. So yeah, change what you need to.

lucianopaz:
I wouldn't add them as `Deterministics`. The executors are used in different contexts for different purposes. The transformed executor is used in `sample` and in `sample_posterior_predictive`, and the latter relies on distributions being in untransformed, transformed or observed values in the sampling state, but never in deterministics. I can tackle this issue later this week but I would do the following:

1. The deterministic values are collected with a callback that does a forward pass through the model, so it also collects the untransformed distribution values. I would simply change this callback to also return the relevant subset of untransformed variables that should be included into the trace.
2. To know the list of distributions that should be stored, we could either rely on the test forward pass that we do in build logp function, or we could add an attribute to the sampling state, that stores the names of the autotransformed distributions.

ColCarroll:
Thanks, @lucianopaz and @rrkarim!

I think I got to the same place as you are suggesting with 1. I am just finding it hard to figure out where to add that collection logic! :grin: 

I am relying on the nice `pymc4.utils.NameParts.is_transformed` to decide if a distribution should be stored.

PR #190, opened 2019-12-18
Committing a basic usage notebook

ericmjl:
This was partially a tutorial for myself, and also a way to introduce a "simple complex example" for PyMC3 users.

Also discovered some UI/UX limitations, and will be posting them shortly.


lucianopaz:
Nice! Thanks @ericmj! I would wait for #180 so you don't have to plot the log off the scale. One other thing, could you try to add a cell on posterior predictive and prior predictive sampling?

lucianopaz:
One more super small nitpick that could cause confusion further down the line, `yield` doesn't return a probability distribution, it returns a value that is drawn or is assumed to come from the probability distribution.

You can try this:
```python
@pm.model
def model():
    dist = pm.Normal("x", 0, 1)
    x = yield dist
    print(dist)
    print(x)

_, state = pm.evaluate_model(model())
```

You'll see that it will print out that `dist` is our `Normal` distribution instance while `x` will be a tensor of values. Using the trick of not yielding from the distribution immediately, you can modify the distribution inside the model's function before you draw samples from it. For example, you could collect the distribution's `log_prob` in a deterministic like this:

```python
@pm.model
def model():
    dist = pm.Normal("x", 0, 1)
    x = yield dist
    logp = pm.Deterministic("logp", dist.log_prob(x))

_, state = pm.evaluate_model(model())
```

You can also play with the distribution's shape and I imagine that you could try many other things that don't come to my mind now.

I know that what I'm saying isn't basic usage and shouldn't be in this guide that you're writing, but I think that it's important not to mix the terms distribution, random variable and samples from a distribution.

ferrine:
<p>I'm interested in untransformed sigma, it should be visible in the chain. If not, it is a case we should fix that</p>

notebooks/basic-usage.ipynb
```diff
@@ -0,0 +1,310 @@
+{
```
twiecki: <p>Should always specify whether we talk about PyMC3 or PyMC4 here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

twiecki: <p>I think we should start putting the <code>name</code> kwarg as the last one everyone.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

twiecki: <p>I think we moved this to the top namespace now.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

twiecki: <p>We should probably change that behavior and just return an arviz trace object with no reference to the tensors.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

twiecki: <p>if you add a <code>;</code> at the end you suppress the weird plot return value being printed in the ouptut.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

lucianopaz:
> 
> 
> I'm interested in untransformed sigma, it should be visible in the chain. If not, it is a case we should fix that

@ferrine, it's currently an open issue (#169) and @ColCarroll was working on a PR to fix it (#180)

notebooks/basic-usage.ipynb
ericmjl: <p>@twiecki I think I have made that distinction clear here?</p>

ericmjl: <p>Yeah, my bad. I have used pm.sample correctly, but will remove the import.</p>

ericmjl: <p>That probably should be a later PR, out of scope for this one?</p>

ericmjl: <p>:thumbsup: (I wish there were emojis here)</p>

ericmjl:
> One more super small nitpick that could cause confusion further down the line, yield doesn't return a probability distribution, it returns a value that is drawn or is assumed to come from the probability distribution.

Cool, thanks for the catch! I was definitely being sloppy with terminology there.

notebooks/basic-usage.ipynb
ericmjl: Done!

ericmjl:
> In other words, we should explain that a pm.model is a program that defers its computation and returns control flow back to the PyMC inference engine every time we yield. Perhaps this explanation belongs in a separate notebook?

@eigenfoo, I think @ferrine or someone else explained what's going on in the PyMC design guide notebook - happy to be corrected on this one.

ericmjl:
Any further comments on the notebook? (I think I've addressed all of them.) If not, I'll hit the big green button tonight (Boston time).

twiecki:
Hit it!

notebooks/basic-usage.ipynb
```diff
@@ -0,0 +1,386 @@
...
```
aloctavodia: <p>Probably "sampling stats" is a better description than "diagnostics"</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

aloctavodia: <p>nitpick, what about using a number of samples other than 1000, just to get a more visual understanding of the meaning of each dimension?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc4/pull/190/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

lucianopaz:
@ericmjl, I would wait for #180 to get merged before merging this PR. Once #180 gets in, you wouldn't have to plot the transformed scale, and you would be able to plot the standard `HalfNormal`'s posterior.

ericmjl:
@lucianopaz I’m going to merge this first, so that I can get this out of my headspace, but I’ll also raise an appropriate issue to remind myself to fix this nb once #180 is done.

PR #195, opened 2019-12-19
clean 8 schools example

### LINKED ISSUE #184: eight_schools notebook not working
Cell 9 fails 

tf_trace[1].numpy().shape

AttributeError                            Traceback (most recent call last)
<ipython-input-9-1291ba1b6d25> in <module>
----> 1 tf_trace[1].numpy().shape

AttributeError: 'dict' object has no attribute 'numpy'

aloctavodia:
* fix #184
* add `trace_to_arviz` to pymc4 top-level namespace
* update `trace_to_arviz` to do not include data as part of the posterior, also remove the model name from the variable name (assuming "/" separates the model name from the variable names).

I just clean the example I am not sure if this should be extended to discuss the divergences as in PyMC3 documentation or leave it as a minimal example (probably not very useful).


aloctavodia:
BTW, shouldn't we apply `trace_to_arviz` internally? Or, do we have any good reason for `pm.sample` to return TF tensors and sample_stats instead of and ArviZ's `InferenceData` object?

pymc4/inference/utils.py
```diff
@@ -38,13 +39,15 @@ def trace_to_arviz(pm4_trace, pm4_sample_stats):
...
 
     Returns
     -------
     arviz.data.inference_data.InferenceData
     """
-    import arviz as az
 
-    posterior = {k: np.swapaxes(v.numpy(), 1, 0) for k, v in pm4_trace.items()}
+    posterior = {
+        k.split("/")[1]: np.swapaxes(v.numpy(), 1, 0) for k, v in pm4_trace.items() if "/" in k
```
lucianopaz: This split wouldn't be safe for variables in nested models. Those would have their names as: `outer_model/inner_model/var_name`. Why don't we leave the name with the entire scope? Does arviz complain about the slash characters?

aloctavodia: yeah, I was not sure this was a good idea :-) I just wanted to remove it because I think is kind of redundant information. We can leave it. Or we can do something like str.split('/', 1)[1].
ArviZ does not complaint about slash characters

lucianopaz: If arviz doesn't complain about the slash, I would just leave the entire name for now. The full name is needed when one wants to fix values in forward sampling

twiecki:
> BTW, shouldn't we apply `trace_to_arviz` internally? Or, do we have any good reason for `pm.sample` to return TF tensors and sample_stats instead of and ArviZ's `InferenceData` object?

I agree we should do that.

aloctavodia:
Should we merge this, or should I keep working?

lucianopaz:
Let's merge this

PR #204, opened 2019-12-29
Load data function added to utils

fonnesbeck:
Ported from PyMC3. Changed notebooks to use the function.

Also added seaborn to requirements.txt, since at least one notebook uses it.


ColCarroll:
Thanks -- `seaborn` we can actually remove (I removed all those plots from the rugby example), but that can be a separate pr.

Issue #197, opened 2019-12-20
Implement Mixture Distribution


lucianopaz:
This is linked to #44. At the moment we don't have a working `Mixture` distribution. I'm opening this as a smaller (kinda-) begginer friendly issue. It should work on top of [`tfd.Mixture`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Mixture) but be able to take a list of pymc4 `Distribution`s (not values yielded from distributions!!) as its components.

rrkarim:
There was an issue on tf side, tensorflow/probability#684, I've opened a PR for it but still waiting for some response on it.

golanor:
Hi,

I'd like to work on this issue. From the thread on tfp it seems like the issue there is fixed.

rrkarim:
@golanor the issue hasn't been solved yet, unfortunately. Here is the PR for it #209 and the discussion of the `tf` issue: tensorflow/tensorflow/issues/35099.

TomKealy:
I would also like to work on this issue!

as135:
I'm going to take a look at this and see what I can do!

lucianopaz:
Closed by #284

PR #209, opened 2020-01-13
Add mixture/normal_mixture

rrkarim:
#197
I wasn't able to run sampling with xla enabled, so the tests are pretty trivial for now since the sampling with nuts takes forever to give some results without xla.

I will tackle (but maybe somebody can help me) the problem so flagging as in progress.

[UPD] my fault, tfp versioning problem, in master the bug was fixed recently.


ColCarroll:
I think also you also need to add a line in `distributions/__init__.py`:

```python
from .mixture import *
```

the tests look quite good, though, thank you!

rrkarim:
Concern about the naming of transforms, if somebody is gonna use `pm.transforms.Invert(pm.transforms.Transform(...))` it is not gonna work with executor, since it can detect its `scoped_name`. I've added name to `pm.tranforms.Invert` but now executor doesn't see full scoped name.

rrkarim:
It seems like the PR for tensorflow fix was pulled back because of forward compatibility. The tests shouldn't pass.

twiecki:
Seems like tests fail.

rrkarim:
> Seems like tests fail.

It is a tf side issue, tensorflow/tensorflow/issues/35099.

junpenglao:
I pinged the thread to hopefully get some help.

rrkarim:
@junpenglao can we ping the thread again? I wasn't able to get the response.

kyleabeauchamp:
I pinged the TF thread as I tried to do some modeling in pymc4 and also hit the same bug.

rrkarim:
@junpenglao since the tests are not that great for now, can I just mark tests with `batch_ndim=0` to `xfail`? Then hopefully change it back once the issue solved on the tf side.

PR #215, opened 2020-02-04
Add default transform(sigmoid) for UnitContinuousDistribution

### LINKED ISSUE #214: UnitContinuousDistribution does not init the default transform
I was looking a bit into the transforms in pymc4 after a discussion about transforms in pymc3 and I noticed that the `UnitContinuousDistribution` does not implement its default transform (we do this for the `PositiveContinuousDistribution`).

This should be simple to solve. We should just copy what we did for the `PositiveContinuousDistribution` but use the `Sigmoid` bijector instead of the `Log` by default.

Sayam753:
Fixes #214 
 - Added _init_transform method for UnitContinuousDistribution
 - Added Sigmoid bijector in transforms.py

### DIFF
--- a/pymc4/distributions/distribution.py
+++ b/pymc4/distributions/distribution.py
@@ -207,6 +207,12 @@ class BoundedContinuousDistribution(ContinuousDistribution, BoundedDistribution)
 
 
 class UnitContinuousDistribution(BoundedContinuousDistribution):
+    def _init_transform(self, transform):
+        if transform is None:
+            return transforms.Sigmoid()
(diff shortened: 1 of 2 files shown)


lucianopaz:
Fantastic @Sayam753! Thanks a lot for the quick fix! I'll merge this as it is, but the code coverage failure means that we are not testing `pm.sample` on a model that has a `UnitContinuousDistribution` (for example, a `Uniform`). In a later PR we can add a test in `test_sampling.py` that does that.

PR #218, opened 2020-02-05
ENH: Add BetaBinomial Distribution

tirthasheshpatel:
Addresses #44.

Summary of my changes:
 - Add ``BoundedDiscreteCompoundDistribution`` ABC to be consistent with the API.
 - Add BetaBinomial Compound Discrete Distribution.

### DIFF
--- a/pymc4/distributions/discrete.py
+++ b/pymc4/distributions/discrete.py
@@ -4,11 +4,13 @@ from tensorflow_probability import distributions as tfd
 from pymc4.distributions.distribution import (
     PositiveDiscreteDistribution,
     BoundedDiscreteDistribution,
+    BoundedDiscreteCompoundDistribution,
 )
 
 __all__ = [
     "Bernoulli",
     "Binomial",
+    "BetaBinomial",
     "DiscreteUniform",
     "Categorical",
     "Geometric",
@@ -126,6 +128,74 @@ class Binomial(BoundedDiscreteDistribution):
 
     def upper_limit(self):
         return self.conditions["total_count"]
+
+
+class BetaBinomial(BoundedDiscreteCompoundDistribution):
+    r"""Bounded Discrete compound Beta-Binomial Random Variable
+
+    The pmf of this distribution is
+
+    .. math:: f(x \mid n, \alpha, \beta) = n \choose x \frac{B(x + \alpha, n - x + \beta)}{B(\alpha, \beta)}
+
+    .. plot::
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import scipy.stats as st
+        plt.style.use('seaborn-darkgrid')
+        lows = [1, -2]
+        highs = [6, 2]
+        for low, high in zip(lows, highs):
+            x = np.arange(low, high+1)
+            pmf = [1 / (high - low)] * len(x)
+            plt.plot(x, pmf, '-o', label='low = {}, high = {}'.format(low, high))
+        plt.xlabel('x', fontsize=12)
+        plt.ylabel('f(x)', fontsize=12)
+        plt.ylim(0, 0.4)
+        plt.legend(loc=1)
+        plt.show()
+
+    ========  ===============================================
+    Support   :math:`x \in {0, 1 + 1, \ldots, n}`
+    Mean      :math:`\dfrac{n \alpha}{\alpha + \beta}`
+    Variance  :math:`\dfrac{n \alpha \beta (\alpha + \beta + n)}{(\alpha + \beta)^2 (\alpha + \beta + 1)}`
+    ========  ===============================================
+
+    Parameters
+    ----------
+    total_count : int
+        Number of trials `n` (total_count>=0)
+    concentration0 : float
+        `alpha` parameter of the Beta Distribution (concentration0 > 0)
+    concentration1 : float
+        `beta` parameter of the Beta Distribution (concentration1 > 0)
+    """
+
+    def __init__(self, name, total_count, concentration0, concentration1, **kwargs):
+        super().__init__(
+            name,
+            total_count=total_count,
+            concentration0=concentration0,
+            concentration1=concentration1,
+            **kwargs,
+        )
+
+    @staticmethod
+    def _init_distribution(conditions):
+        total_count, concentration0, concentration1 = (
+            conditions["total_count"],
+            conditions["concentration0"],
+            conditions["concentration1"],
+        )
+        return tfd.DirichletMultinomial(
+            total_count=total_count, concentration=[concentration0, concentration1]
+        )
+
+    def lower_limit(self):
+        return 0
+
+    def upper_limit(self):
+        return self.conditions["total_counts"]
 
 
 class DiscreteUniform(BoundedDiscreteDistribution):
--- a/pymc4/distributions/discrete.py
+++ b/pymc4/distributions/discrete.py
@@ -4,7 +4,6 @@ from tensorflow_probability import distributions as tfd
 from pymc4.distributions.distribution import (
     PositiveDiscreteDistribution,
     BoundedDiscreteDistribution,
-    BoundedDiscreteCompoundDistribution,
 )
 
 __all__ = [
@@ -130,12 +129,16 @@ class Binomial(BoundedDiscreteDistribution):
         return self.conditions["total_count"]
 
 
-class BetaBinomial(BoundedDiscreteCompoundDistribution):
+class BetaBinomial(BoundedDiscreteDistribution):
     r"""Bounded Discrete compound Beta-Binomial Random Variable
 
     The pmf of this distribution is
 
     .. math:: f(x \mid n, \alpha, \beta) = n \choose x \frac{B(x + \alpha, n - x + \beta)}{B(\alpha, \beta)}
+
+    where :math:`n` = ``total_count``
+          :math:`\alpha` = ``concentration0``
+          :math:`\beta` = ``concentration1``
 
     .. plot::
 
@@ -167,9 +170,9 @@ class BetaBinomial(BoundedDiscreteCompoundDistribution):
     total_count : int
         Number of trials `n` (total_count>=0)
     concentration0 : float
-        `alpha` parameter of the Beta Distribution (concentration0 > 0)
+        :math:`\alpha` parameter of the Beta Distribution (concentration0 > 0)
     concentration1 : float
-        `beta` parameter of the Beta Distribution (concentration1 > 0)
+        :math:`\beta` parameter of the Beta Distribution (concentration1 > 0)
     """
 
     def __init__(self, name, total_count, concentration0, concentration1, **kwargs):
--- a/pymc4/distributions/discrete.py
+++ b/pymc4/distributions/discrete.py
@@ -143,11 +143,12 @@ class BetaBinomial(BoundedDiscreteCompoundDistribution):
         import numpy as np
         import scipy.stats as st
         plt.style.use('seaborn-darkgrid')
-        lows = [1, -2]
-        highs = [6, 2]
-        for low, high in zip(lows, highs):
-            x = np.arange(low, high+1)
-            pmf = [1 / (high - low)] * len(x)
+        n = 10
+        x = np.arange(0, n)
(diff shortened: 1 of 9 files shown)


pymc4/distributions/discrete.py
```diff
@@ -128,6 +130,74 @@ def upper_limit(self):
...
+
+    .. math:: f(x \mid n, \alpha, \beta) = n \choose x \frac{B(x + \alpha, n - x + \beta)}{B(\alpha, \beta)}
+
+    .. plot::
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import scipy.stats as st
+        plt.style.use('seaborn-darkgrid')
+        lows = [1, -2]
```
tirthasheshpatel: Oh, shoot! Adopted docstring from uniform distribution but forgot to edit it! Let me do that real quick

pymc4/distributions/distribution.py
```diff
@@ -202,6 +203,10 @@ class BoundedDiscreteDistribution(DiscreteDistribution, BoundedDistribution):
     ...
 
 
+class BoundedDiscreteCompoundDistribution(BoundedDiscreteDistribution):
```
lucianopaz: Could you explain why we should add the `BoundedDiscreteCompoundDistribution` distribution and not use `BoundedDiscreteDistribution` directly?

pymc4/distributions/discrete.py
```diff
@@ -128,6 +130,75 @@ def upper_limit(self):
         return self.conditions["total_count"]
 
 
+class BetaBinomial(BoundedDiscreteCompoundDistribution):
+    r"""Bounded Discrete compound Beta-Binomial Random Variable
+
+    The pmf of this distribution is
+
+    .. math:: f(x \mid n, \alpha, \beta) = n \choose x \frac{B(x + \alpha, n - x + \beta)}{B(\alpha, \beta)}
```
lucianopaz: Thanks for copying over the docstring from pymc3, but we have a parameter name discrepancy. Here, we use `concentration` instead of alpha and beta, and `total_count` instead of n. People will be confused by the equation not matching the function's signature. I'm not sure what's best to do here. We can either change the parameter names in the equation or we could add a table after the equation that indicates the correspondence between the equation variables and the parameters in the class' signature.

pymc4/distributions/distribution.py
tirthasheshpatel: If we add other compound distributions in the future than this would be a more interpretable parent class. But it's not necessary! If you want, I can remove it...

pymc4/distributions/discrete.py
tirthasheshpatel: I believe that would make the docstring un-necessarily long but we can just add a docstring saying ``n = total_counts, \alpha = concentration0 and \beta = concentration1``.

tirthasheshpatel:
I will add tests in some time! Meanwhile, do these samples make sense:
```python
>>> samples.posterior['bbmodel/x']
<xarray.DataArray 'bb_model/x' (chain: 1, draw: 10, bb_model/x_dim_0: 2)>
array([[[ -31.816374,   30.671614],
        [ -34.609715,   33.548725],
        [ -76.14795 ,   75.38489 ],
        [-124.313614,  122.80483 ],
        [-178.72943 ,  177.00676 ],
        [-178.78172 ,  176.54834 ],
        [-177.54681 ,  176.86978 ],
        [-179.86192 ,  178.99353 ],
        [-167.46588 ,  165.29099 ],
        [-167.0702  ,  165.82793 ]]], dtype=float32)
Coordinates:
  * chain             (chain) int32 0
  * draw              (draw) int32 0 1 2 3 4 5 6 7 8 9
  * bb_model/x_dim_0  (bb_model/x_dim_0) int32 0 1
```

I think this must be a single dimension vector like binomial distribution! Am I missing sometinhg?

pymc4/distributions/distribution.py
lucianopaz: My question was more about, what make a _compound distribution_ different from a _distribution_ from a programmers point of view. I understand that they are different mathematical entities, but what does the compoundness change from the `class` inheritance point of view.

pymc4/distributions/discrete.py
lucianopaz: Ok, lets add that, just make sure that sphinx will render the `total_counts` and `concentration*` as [inline code](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup), and the other stuff as [inline math](http://www.sphinx-doc.org/es/stable/ext/math.html)

lucianopaz:
> 
> 
> I will add tests in some time! Meanwhile, do these samples make sense:
> 
> ```python
> >>> samples.posterior['bb_model/x']
> <xarray.DataArray 'bb_model/x' (chain: 1, draw: 10, bb_model/x_dim_0: 2)>
> array([[[ -31.816374,   30.671614],
>         [ -34.609715,   33.548725],
>         [ -76.14795 ,   75.38489 ],
>         [-124.313614,  122.80483 ],
>         [-178.72943 ,  177.00676 ],
>         [-178.78172 ,  176.54834 ],
>         [-177.54681 ,  176.86978 ],
>         [-179.86192 ,  178.99353 ],
>         [-167.46588 ,  165.29099 ],
>         [-167.0702  ,  165.82793 ]]], dtype=float32)
> Coordinates:
>   * chain             (chain) int32 0
>   * draw              (draw) int32 0 1 2 3 4 5 6 7 8 9
>   * bb_model/x_dim_0  (bb_model/x_dim_0) int32 0 1
> ```
> 
> I think this must be a single dimension vector like binomial distribution! Am I missing something?

I would need to see the code with which you generated that. Did you `pm.sample`? Why don't you try to simply do

```python
dist = BetaBinomial(some_parameters)
dist.sample(sample_shape=some_shape)
```

and see if what you get back matches what you'd expect?

pymc4/distributions/distribution.py
tirthasheshpatel: Hmm... You are right! It will mostly not affect anything during coding. I will change that too! Thanks

pymc4/distributions/discrete.py
tirthasheshpatel: On it!

tirthasheshpatel:
> I will add tests in some time! Meanwhile, do these samples make sense:
> 
> ```python
> >>> samples.posterior['bb_model/x']
> <xarray.DataArray 'bb_model/x' (chain: 1, draw: 10, bb_model/x_dim_0: 2)>
> array([[[ -31.816374,   30.671614],
>         [ -34.609715,   33.548725],
>         [ -76.14795 ,   75.38489 ],
>         [-124.313614,  122.80483 ],
>         [-178.72943 ,  177.00676 ],
>         [-178.78172 ,  176.54834 ],
>         [-177.54681 ,  176.86978 ],
>         [-179.86192 ,  178.99353 ],
>         [-167.46588 ,  165.29099 ],
>         [-167.0702  ,  165.82793 ]]], dtype=float32)
> Coordinates:
>   * chain             (chain) int32 0
>   * draw              (draw) int32 0 1 2 3 4 5 6 7 8 9
>   * bb_model/x_dim_0  (bb_model/x_dim_0) int32 0 1
> ```
> 
> I think this must be a single dimension vector like binomial distribution! Am I missing something?


So, I used this code to generate those samples
```python
import pymc4 as pm

@pm.model
def bb_model():
    x = yield pm.BetaBinomial('x', 10, 1., 2.)
    return x

samples = pm.sample(bb_model(), num_samples=10, burn_in=10, num_chains=1)
samples
```

These samples have values like ``[-177.54681,  176.86978]`` that does not make sense to me! I mean the values must be in between 0 and ``total_counts`` and 1-dimensional (by default) as, at last, they are samples from binomial. 

OTOH, when I directly sample from ``BetaBinomial('x', 10, 1., 2.)`` the samples are in between 0 and ``total_counts`` but 2D (which, according to the tfp documentation, is expected). Am I missing something here? There is also no ``BetaBinomial`` distribution in tfp. Can someone guide me through this please?

lucianopaz:
Ok, the problem is that at the moment we have only interfaced `pm.sample` to perform NUTS. This is a form of HMC that is only suited to sample from continuous variables. The second part is that we haven't implemented an auto transformation mechanism for bounded variables in general, and for discrete bounded variables in particular. MCMC usually propose samples outside of the distribution's support, so we map those proposals onto the correct domain. So far we've done this for positive continuous and unit continuous variables only.

The important things that should be tested are that the distribution's `sample` and `log_prob` methods work.

tirthasheshpatel:
Got it @lucianopaz. Thanks!

For now, this should work!

tests/test_distributions.py
```diff
@@ -15,6 +15,15 @@ def random_variable_args():
         ("Bernoulli", {"probs": 0.5, "sample": 1.0}),
         ("Beta", {"concentration0": 1, "concentration1": 1}),
         ("Binomial", {"total_count": 5.0, "probs": 0.5, "sample": 1.0}),
+        (
+            "BetaBinomial",
+            {
+                "total_count": 10.0,
+                "concentration0": 1.0,
+                "concentration1": 2.0,
+                "sample": [3.0, 7.0],
```
lucianopaz: Great! If you want to make sure that things are working properly, you could add a value to the sample list that is outside of the bounds, the distribution `log_prob` should return Nan or -inf in those cases

tirthasheshpatel: I tried it out but ``log_prob`` returns finite values no matter what the sample is!
```python
>>> import pymc4 as pm
>>> dist = pm.BetaBinomial('x', 10, 1., 2.)
2020-02-07 13:29:06.852663: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2
>>> dist.log_prob([3., 10.])
<tf.Tensor: shape=(), dtype=float32, numpy=-9.703817>
>>> dist.log_prob([10., 20.])
<tf.Tensor: shape=(), dtype=float32, numpy=-62.71588>
```

lucianopaz:
That's strange, I'll look at this a bit closer

tirthasheshpatel:
Most probably, it is an error for TensorFlow probability.

tirthasheshpatel:
I think this is an error for many other multivariate distributions:

- Dirichlet
```python
>>> dist = pm.Dirichlet('x', [1.0, 2.0])
>>> dist.log_prob(1.0)
<tf.Tensor: shape=(), dtype=float32, numpy=0.6931472>
>>> dist.log_prob(2.0)
<tf.Tensor: shape=(), dtype=float32, numpy=1.3862944>
>>> dist.log_prob(3.0)
<tf.Tensor: shape=(), dtype=float32, numpy=1.7917595>
```

- Multinomial
```python
>>> dist = pm.Multinomial('x', 10, [0.3, 0.7])
>>> dist.log_prob([3.0, 7.0])
<tf.Tensor: shape=(), dtype=float32, numpy=-1.3211508>
>>> dist.log_prob([3.0, 10.0])
<tf.Tensor: shape=(), dtype=float32, numpy=-8.970427>
```

- VonMisesFisher (this is very suspicious!)
```python
>>> dist = pm.VonMisesFisher('x', [10.0, 10.0], 1.)
>>> dist.log_prob([1.0, 1.0])
<tf.Tensor: shape=(), dtype=float32, numpy=17.926208>
>>> dist.log_prob([10.0, 10.0])
<tf.Tensor: shape=(), dtype=float32, numpy=197.92621>
>>> dist.log_prob([100.0, 100.0])
<tf.Tensor: shape=(), dtype=float32, numpy=1997.9263>
```

junpenglao:
The TFP distribution will not validate the input if you leave `validate_args=False`
```
tfd.Dirichlet(np.ones(2)/2, validate_args=Flase).log_prob([.01, .09])  # ==> runs fine
tfd.Dirichlet(np.ones(2)/2, validate_args=True).log_prob([.01, .09])  # ==>  raise error

```

tirthasheshpatel:
Oh! My bad. Will add that too! Thanks @junpenglao

lucianopaz:
Thanks @tirthasheshpatel!

Issue #167, opened 2019-12-01
Over reliance on evaluate_model


lucianopaz:
Currently, `sample`, `sample_prior_predictive` and `sample_posterior_predictive` all use a first call to some variant of `evaluate_model` just to get some meta information of the model:
* Identify the variables that define the sampling state, along with their shape (in `sample`)
* Identify the distribution shape and the observed variables (in both forward sampling methods)
* To help with auto batching (in both forward sampling methods)

All of these will lead to problems for models that cannot generate samples from the prior. For example, models with `Flat` priors (see [this pymc3 issue](https://github.com/pymc-devs/pymc3/issues/3294))

I think that the solution to this problem is to introduce some purely symbolic model parser to get the metadata we need. I think that we should do this with [symbolic-pymc](https://github.com/pymc-devs/symbolic-pymc)

lucianopaz:
Pinging @brandonwillard over this.

If we don't go symbolic, the alternatives I can think of are:

1. making a special executor that evaluates the model but instead of calling `sample` on distributions, it returns an array of Nan's with the appropriate tensor shape.
2. Doing something that parses the AST to extract the metadata.

I think that option 1 would be easy to implement, but I think that pymc4 could benefit much more by starting to integrate the symbolic-pymc manipulations @brandonwillard is working on.

brandonwillard:
We could almost certainly extract all the necessary information from TFP-generated graphs (which would be much easier with PyMC4/TFP name correspondence, if that's working now).

Otherwise, we could build and use `symbolic-pymc` meta graphs from the start.  If you don't use logic variables with them, they're basically just low-level, generic delayed TF graphs&mdash;that are easily converted to real TF graphs with `MetaSymbol.reify`.

I'll need to know more about the shape-related needs, because there may be TF-only difficulties with obtaining shape information for a specific node in the exact  form/way we need it.

lucianopaz:
@brandonwillard, how would we bypass eager mode to get the tf graph? Would we need a `function` decorator? That would still have to rely on `evaluate_model` to trace the function and generate the graph, right?

As for the shape issues, tfp has the three semantic shapes: sample, batch and event. Currently, all three methods always leave
[...]

PR #219, opened 2020-02-08
ENH: add Flat and HalfFlat Distributions

tirthasheshpatel:
Addresses #44.

Summary of my changes:
 - Add ``Flat`` distribution
 - Add ``HalfFlat`` distribution

### Question

Do these need tests?


pymc4/distributions/continuous.py
```diff
@@ -1165,6 +1169,51 @@ def upper_limit(self):
...
+
+    def __init__(self, name, **kwargs):
+        super().__init__(name, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions):
+        return tfd.Uniform(low=0.0, high=np.inf)
+
+    def log_prob(self, value):
+        return tf.cond(
```
brianwa84: You should use tf.where here

lucianopaz:
@tirthasheshpatel, thanks again. Contrary to what it may look like, these two distributions will need more tests than the rest for two big reasons:

1. They are improper distributions, so we can't draw forward samples from them (`Flat.sample`). I'm my opinion, if someone tries to call their `sample` method, a `TypeError` should be raised. However, we should be able to do inference in models that have flat or half flat priors (`pm.sample`), and given their posterior samples, we should be able to call sample posterior predictive on them. So they must behave a bit different from the rest of our distributions.
2. There is no tfp equivalent, so we must test that the returned `log_prob` values are correct. That sounds simple because the log_prob should always be zero. The difficulty here is the shape handling. If the `value` passed to `log_prob` has a rank that is lower than the distribution's event shape, we should raise an error, if the value has a shape that doesn't broadcast with the distribution's batch+event shape, we should raise an error. Actually, I think it's even more stringent than that, the rightmost axis of the passed values should exactly match the distribution's event shape, not just broadcast with it, or an error should be raised. Furthermore, the log prob should sum reduce over the event shape axis, in order to only get back a batch of zeroes.

Point 1, cannot be done until we fix #167, and point 2 is a pit in shape hell, so for now I think that we should leave this PR open and when we fix up #167 we can come back to it.

tirthasheshpatel:
@lucianopaz Sorry for the late reply and thanks for such a detailed review! I will think more about the "shape hell" and try to raise relevant errors. Meanwhile, I will also try to catch up with #167 and try to resolve it!

lucianopaz:
@tirthasheshpatel, for now don't worry about #167. I'm working on it in [this branch](https://github.com/lucianopaz/pymc4/tree/test_point), but it requires #193 to be finished first. My recomendation is to do one of the following:

1. Continue to implement regular distributions like what you did with the `BetaBinomial`, and at a later time we can come back to this PR and finish it up.
2. Write the tests for the `Flat` and `HalfFlat` distributions, make sure that the `log_prob` outputs the correct shape, and then mark the tests for predictive sampling and `pm.sample` as expected failures on pytest. Once we finish with #167, we'll remove the expected failure mark and ensure that these tests pass.

pymc4/distributions/continuous.py
tirthasheshpatel: Done. Thanks!

tirthasheshpatel:
Some distributions in #44 don't have a tfp equivalent. So, do we have to wait for it to be implemented or is there a way around? (I was thinking about implementing classes inheriting from tfp's ``Distribution`` class)

pymc4/distributions/continuous.py
```diff
@@ -1165,6 +1169,103 @@ def upper_limit(self):
...
+        value = tf.convert_to_tensor(value)
+        # if not tf.rank(value):
+        #     value = tf.convert_to_tensor([value])
+        # value_shape = value.shape
+        # # if the rank of the `value` is less than distribution's `event shape` raise an error
+        # if tf.rank(value) < tf.rank(self._distribution.event_shape):
+        #     raise ValueError("Rank of input tensor less than distribution's event shape")
+        # # if the rightmost axis of `value` doesn't match the distribution's `event_shape`, raise an error
+        # if (
+        #     len(self._distribution.event_shape)
```
tirthasheshpatel: > 2. There is no tfp equivalent, so we must test that the returned `log_prob` values are correct. That sounds simple because the log_prob should always be zero. The difficulty here is shape handling. If the `value` passed to `log_prob` has a rank that is lower than the distribution's event shape, we should raise an error, if the value has a shape that doesn't broadcast with the distribution's batch+event shape, we should raise an error. Actually, I think it's even more stringent than that, the rightmost axis of the passed values should exactly match the distribution's event shape, not just broadcast with it, or an error should be raised. Furthermore, the log prob should sum reduce over the event shape axis, in order to only get back a batch of zeroes.

As the distribution is univariate the event_shape and batch_shape will always be ``()`` and so I think we don't need to worry about it, right?

pymc4/distributions/continuous.py
```diff
@@ -1165,6 +1169,103 @@ def upper_limit(self):
...
+        #     )
+        # # if the shape of `value` doesn't match the distribution's `batch_shape`, raise an error
+        # if (
+        #     len(self._distribution.batch_shape)
+        #     and value_shape[-2] != self._distribution.batch_shape[-1]
+        # ):
+        #     raise ValueError(
+        #         "Batch shape of input tensor not consistent with the distributions's batch shape"
+        #     )
+        if tf.rank(value) > 2:
```
tirthasheshpatel: I have reduced sum when the shape of value if ``(n_samples, batch_shape, event_shape)`` along the ``event_shape`` axis. Shouldn't we reduce sum over ``n_samples`` axis?

pymc4/distributions/continuous.py
lucianopaz: The defaults will be `()`, but we can stack multiple independent variable that follow this distribution with something similar to `tfd.Sample`, and that will lead to an event shape that could be a tuple of any length.
Furthermore, in #167, we are also aiming to provide a similar mechanism as `tfd.Sample` but instead of stacking events, we will stack batches. So in the end, any distribution can have any number of event and batch axes. I just realized that I wrote the singular axis in my previous comment, but I meant it to be the plural, axes.

pymc4/distributions/continuous.py
lucianopaz: We only must reduce sum over the event shape axes. That's what truly distinguishes between the batch and event shapes. For example,

```python
>>> import tensorflow as tf
>>> from tensorflow_probability import distributions as tfd
>>> 
>>> d = tfd.Normal(loc=tf.zeros(1, 2), scale=1)
>>> d = tfd.Sample(d, sample_shape=(3, 4, 5))
>>> d.batch_shape.as_list()
[1, 2]
>>> d.event_shape.as_list()
[3, 4, 5]
>>> 
>>> x = d.sample(sample_shape=(6, 7))
>>> x.numpy().shape
(6, 7, 1, 2, 3, 4, 5)
>>> v = d.log_prob(x)
>>> v.numpy().shape
(6, 7, 1, 2)
```

pymc4/distributions/continuous.py
```diff
@@ -1165,6 +1169,103 @@ def upper_limit(self):
...
+
+    def __init__(self, name, **kwargs):
+        super().__init__(name, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions):
+        return tfd.Uniform(low=-np.inf, high=np.inf)
+
+    def log_prob(self, value):
+        # convert the value to tensor
```
lucianopaz: I think that you're doing too many if statements. I recommend that you start out with `expected = tf.zeros(self._distribution.batch_shape + self._distribution.event_shape)`. This would be the expected shape of a call to a `tfd.Distribution.sample()`.

You should first check if `value` has a matching event shape part to `expected`, if yes then broadcast `expected` and `value` and create a zeros tensor with the resulting shape. Finally, you'll have to reduce sum over all event shape axes.

pymc4/distributions/continuous.py
tirthasheshpatel: Oh! This is very clear now. Thanks!!

lucianopaz:
Pinging @tirthasheshpatel, just to let you know that I've opened #227. Once that gets merged in, you should merge it into your branch and you'll be able to run all of the tests required by the `Flat` and `HalfFlat` distributions.

tirthasheshpatel:
@lucianopaz sorry for getting back to you late. I will complete this by today.

tests/test_distributions.py
```diff
@@ -348,6 +362,8 @@ def test_rvs_test_point_are_valid(tf_seed, distribution_conditions):
     dist_class = getattr(pm, distribution_name)
     dist = dist_class(name=distribution_name, **conditions)
     test_value = dist.test_value
+    if distribution_name in ["Flat", "HalfFlat"]:
+        pytest.skip("Flat and HalfFlat distributions don't support sampling.")
```
lucianopaz: We shouldn't skip this test. In the case of the `Flat` and `HalfFlat`, instead of comparing the `test_value.shape` against the `test_sample.shape`, we should compare it against the `sample_shape + batch_shape + event_shape` that we expect to get.

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,84 @@ def upper_limit(self):
...
+        value = tf.convert_to_tensor(value)
+        expected = tf.zeros(self._distribution.batch_shape + self._distribution.event_shape)
+        # check if the event shape matches
+        if (
+            len(self._distribution.event_shape)
+            and value.shape[-len(self._distribution.event_shape) :]
+            != self._distribution.event_shape
+        ):
+            raise ValueError("values not consistent with the event shape of distribution")
```
lucianopaz: You should write a test that raises this exception to make sure it's working

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,84 @@ def upper_limit(self):
...
+        # broadcast expected to shape of value
+        try:
+            expected = tf.broadcast_to(expected, value.shape)
+        except tf.python.framework.errors_impl.InvalidArgumentError:
+            raise ValueError("value can't be broadcasted to expected shape")
```
lucianopaz: You should write a test that raises this exception to make sure it's working

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,84 @@ def upper_limit(self):
...
+    def _init_distribution(conditions):
+        return tfd.Uniform(low=-np.inf, high=np.inf)
+
+    def log_prob(self, value):
+        # convert the value to tensor
+        value = tf.convert_to_tensor(value)
+        expected = tf.zeros(self._distribution.batch_shape + self._distribution.event_shape)
+        # check if the event shape matches
+        if (
+            len(self._distribution.event_shape)
```
lucianopaz: A minor nitpick, our distributions now have the `event_shape` and `batch_shape` properties, so you don't have to access the `_distribution` attribute to look them up. You can do `self.event_shape` and `self.batch_shape` instead

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,84 @@ def upper_limit(self):
...
+        value = tf.convert_to_tensor(value)
+        value = tf.where(value > 0, x=0.0, y=-np.inf)
+        expected = tf.zeros(self._distribution.batch_shape + self._distribution.event_shape)
+        # check if the event shape matches
+        if (
+            len(self._distribution.event_shape)
+            and value.shape[-len(self._distribution.event_shape) :]
+            != self._distribution.event_shape
+        ):
+            raise ValueError("values not consistent with the event shape of distribution")
```
lucianopaz: The same comment as for the `Flat` distribution, we need a test that raises this exception

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,84 @@ def upper_limit(self):
...
+        # broadcast expected to shape of value
+        try:
+            expected = tf.broadcast_to(expected, value.shape) + value
+        except tf.python.framework.errors_impl.InvalidArgumentError:
+            raise ValueError("value can't be broadcasted to expected shape")
```
lucianopaz: The same comment as for the `Flat` distribution, we need a test that raises this exception

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,84 @@ def upper_limit(self):
...
+        # check if the event shape matches
+        if (
+            len(self._distribution.event_shape)
+            and value.shape[-len(self._distribution.event_shape) :]
+            != self._distribution.event_shape
+        ):
+            raise ValueError("values not consistent with the event shape of distribution")
+        # broadcast expected to shape of value
+        try:
+            expected = tf.broadcast_to(expected, value.shape)
```
lucianopaz: Does this work if `value.shape=(3,)` but the distribution has `batch_shape=(4,)` and `event_shape=(3,)`, so the expected shape is `(4, 3)`? Wont it raise an error for that kind of situation?

tirthasheshpatel: Yes. I think this can be fixed by checking ``tf.rank(values) < tf.rank(self.batch_shape + self.event_shape)``. If this condition evaluates to ``True``,  we only check if the values are consistent with ``batch_shape`` and we don't need to broadcast. While if the condition is ``False``,  let ``tf.broadcast_to`` handle it. What do you say? Is there a way around?

<details>
<summary>Code</summary>

```python
if tf.rank(value) < tf.rank(self.batch_shape + self.event_shape):
    if value.shape[:-len(self.event_shape)] != self.batch_shape[-1:-(len(value.shape)-len(self.event_shape))]:
        raise ValueError("batch shape of values is not consistent with distribution's batch_shape")
else:
    try:
        ...
```

</details>

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,97 @@ def upper_limit(self):
...
+        # convert the value to tensor
+        value = tf.convert_to_tensor(value)
+        expected = tf.zeros(self.batch_shape + self.event_shape)
+        # check if the event shape matches
+        if len(self.event_shape) and value.shape[-len(self.event_shape) :] != self.event_shape:
+            raise ValueError("values not consistent with the event shape of distribution")
+        # broadcast expected to shape of value
+        if len(value.shape) < len(self.batch_shape + self.event_shape):
+            if (
+                value.shape[: -len(self.event_shape)]
```
lucianopaz: This won't work if the event shape is a scalar, you'll end up with `value.shape[:0]`. You will have to change this to `value.shape[:len(value.shape) - len(event_shape)]`

pymc4/distributions/continuous.py
```diff
@@ -1174,6 +1177,97 @@ def upper_limit(self):
...
+                != list(reversed(self.batch_shape))[: (len(value.shape) - len(self.event_shape))]
```
lucianopaz: Why do you need to reverse this?

tests/test_distributions.py
```diff
@@ -348,10 +363,23 @@ def test_rvs_test_point_are_valid(tf_seed, distribution_conditions):
...
     logp = dist.log_prob(test_value).numpy()
     assert test_value.shape == test_sample.shape
     assert tuple(test_value.shape.as_list()) == tuple(
         (dist.batch_shape + dist.event_shape).as_list()
     )
     assert not (np.any(np.isinf(logp)) or np.any(np.isnan(logp)))
+
+
+@pytest.mark.parametrize("distribution_name", ["Flat", "HalfFlat"])
+@pytest.mark.parametrize("sample", [tf.zeros(1), tf.zeros((1, 3, 4)), tf.zeros((1, 5, 3, 4))])
```
lucianopaz: We are using fixtures instead of mark parameterize. Could you change it to that?

pymc4/distributions/continuous.py
tirthasheshpatel: If the ``batch_shape`` is ``(1, 2)``, ``event_shape`` is ``(3, 4)``, and we want to sample ``(2, 3, 4)`` then the target must broadcast to ``(1, 2, 3, 4)``. So, we need to check the values of batch_shape from the rightmost axis.

tests/test_distributions.py
tirthasheshpatel: Sure!

pymc4/distributions/continuous.py
tirthasheshpatel: Oh, I see the problem! I have to check the ``batch_shape[len(self.batch_shape) - len(value.shape):]`` instead of reversing it.

tirthasheshpatel:
want to finish this up @lucianopaz?

lucianopaz:
Sorry @tirthasheshpatel for not being able to finish this up sooner. Covid quarantine and all, I can't put almost any time to review pymc stuff. I'll merge this PR because it seems ready to go. Great job!

PR #224, opened 2020-02-14
Added distribution test values

lucianopaz:
This PR partially addresses #167. It only adds test values to every distribution. I'm leaving the implementation of a meta-executor (which would eventually allow us to sample from models that have `Flat` and `HalfFlat` priors) for a future PR.


tests/test_distributions.py
```diff
@@ -1,116 +1,344 @@
...
-    assert x / y is not None
-    assert x // y is not None
-    assert x % y is not None
-    assert x ** y is not None
-    assert -x is not None
+    dist_class = getattr(pm, distribution_name)
+    dist = dist_class(name=distribution_name, **conditions)
+    test_value = dist.test_value
+    test_sample = dist.sample()
+    print(test_value, test_sample)
```
ColCarroll: spare print?
```suggestion
```

PR #225, opened 2020-02-15
TST: add tests for BetaBinomial distribution

tirthasheshpatel:
#224  forgot to add test for BetaBinomial distribution

Here are my changes:
 - Add test for BetaBinomial distribution
 - Concatenate ``concentration0`` and ``concentration1`` parameters to preserve the ``batch_shape``
 - Fix the ``upper_limit`` method the was accessing the wrong key.

### DIFF
--- a/pymc4/distributions/discrete.py
+++ b/pymc4/distributions/discrete.py
@@ -195,7 +195,8 @@ class BetaBinomial(BoundedDiscreteDistribution):
             conditions["concentration1"],
         )
         return tfd.DirichletMultinomial(
-            total_count=total_count, concentration=np.concatenate((concentration0, concentration1), axis=-1)
+            total_count=total_count,
+            concentration=np.concatenate((concentration0, concentration1), axis=-1),
         )
 
     def lower_limit(self):
--- a/pymc4/distributions/discrete.py
+++ b/pymc4/distributions/discrete.py
@@ -1,5 +1,6 @@
 """PyMC4 discrete random variables."""
 import tensorflow as tf
+import numpy as np
 from tensorflow_probability import distributions as tfd
 from pymc4.distributions.distribution import (
     PositiveDiscreteDistribution,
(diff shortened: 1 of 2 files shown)


pymc4/distributions/discrete.py
```diff
@@ -192,14 +195,15 @@ def _init_distribution(conditions):
             conditions["concentration1"],
         )
         return tfd.DirichletMultinomial(
-            total_count=total_count, concentration=[concentration0, concentration1]
+            total_count=total_count,
+            concentration=np.concatenate((concentration0, concentration1), axis=-1),
```
tirthasheshpatel: Should we keep ``validate_args=True`` as the shapes are a little tricky to deal with?

pymc4/distributions/discrete.py
```diff
@@ -176,11 +177,13 @@ class BetaBinomial(BoundedDiscreteDistribution):
     """
 
     def __init__(self, name, total_count, concentration0, concentration1, **kwargs):
+        concentration0 = np.array(concentration0)
```
lucianopaz: Don't try to convert to numpy arrays. You should leave them alone or `tf.convert_to_tensor`

pymc4/distributions/discrete.py
```diff
@@ -176,11 +177,13 @@ class BetaBinomial(BoundedDiscreteDistribution):
...
+        concentration1 = np.array(concentration1)
         super().__init__(
             name,
             total_count=total_count,
-            concentration0=concentration0,
-            concentration1=concentration1,
+            concentration0=concentration0.reshape(concentration0.shape + (1,)),
+            concentration1=concentration1.reshape(concentration1.shape + (1,)),
```
lucianopaz: Never reshape under the hood while you instantiate a distribution. This will bring many shape issues down the line

tests/test_distributions.py
```diff
@@ -32,6 +32,20 @@
...
     },
+    "BetaBinomial": {
+        "scalar_parameters": {
+            "total_count": 5.0,
+            "concentration0": 1.0,
+            "concentration1": 2.0,
+            "sample": [3.0, 7.0],
+        },
+        "multidim_parameters": {
+            "total_count": np.array([20, 2, 5.0], dtype="float32").reshape(-1, 1),
```
lucianopaz: Why do you have to reshape this?

pymc4/distributions/discrete.py
lucianopaz: I'm not sure about it yet

tirthasheshpatel:
Closing this as #231 addresses the issue.

PR #226, opened 2020-02-16
Adds Weibull distribution

sshkhr:
I was looking to contribute to PyMC4 and based on #44 I thought I would add the discrete Weibull distribution for a start. When I looked at the source it seems to me that the Weibull distribution is not available either (the comments mention it as a To-Do). 

TFD is switching from Weibull [which was deprecated in January](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Weibull) to [Weibull CDF](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/WeibullCDF) so I thought I'd update PyMC4 accordingly before starting with discrete Weibull. This is my first contrib to PyMC so please let me know if there are any particular tests you'd like me to add and if something needs to be fixed in the PR. Thanks a ton :smiley:


lucianopaz:
Thanks @sshkhr! We greatly appreciate these PRs! To get the tests to pass you have to fix your code's format. Just run `make black` at the project's root and the error you're seeing in the build pipeline will be fixed.

About tests, it would be great to add a line for this distribution in `test_distributions.py`.

sshkhr:
Thanks a ton, @lucianopaz for the feedback. I've added the tests and ran make black and made the commits required for formatting. However, the tests are still failing. When I run `make black` it says:

> All done! ✨ 🍰 ✨
> 29 files left unchanged.

Any ideas what I should be doing next?

lucianopaz:
The formatting is right but the unit tests are failing. You can take a look at the details by clicking on the failing test and then on [view details on azure pipelines](https://dev.azure.com/pymc-devs/pymc4/_build/results?buildId=253&view=logs&jobId=00b6a206-eb91-585b-3d14-cf1a4d7b1970). The error message you are seeing is because the new distribution isn't being imported into the pymc4 module. You can take a look at how we do this for the other distributions and try to copy it for the `Weibull`.

sshkhr:
Thanks again @lucianopaz for pointing me to the azure pipelines. I fixed the issue with importing Weibull to PyMC4. The tests are failing now because when calling sample() with multidimensional parameters, TFP is unable to set the sample size for a transformed distribution with WeibullCDF bijector. I'll look into how it can be fixed or if you have any ideas please let me know. 

Here is a prototype of the error I have reproduced:

```
p_y = tfd.TransformedDistribution(distribution=tfd.Uniform(low=0, high=1), 
                            bijector=bij.Invert(bij.WeibullCDF(scale=[1.0,1.0], concentration=[0.5,1.0])),
                            name="Weibull")

print(p_y)
tfp.distributions.TransformedDistribution("Weibull", batch_shape=[], event_shape=[], dtype=float32)

y_samples = p_y.sample()
```
 On printing p_y expected batch_shape value should be [2] for sample() to work. Instead running sample() throws this ValueError which causes the multidim test to fail:

```
ValueError                                Traceback (most recent call last)
<ipython-input-46-a68f73fe26d6> in <module>
----> 1 y_samples = p_y.sample()

~/anaconda3/lib/python3.7/site-packages/tensorflow_probability/python/distributions/distribution.py in sample(self, sample_shape, seed, name, **kwargs)
    856       samples: a `Tensor` with prepended dimensions `sample_shape`.
    857     """
--> 858     return self._call_sample_n(sample_shape, seed, name, **kwargs)
    859 
    860   def _call_log_prob(self, value, name, **kwargs):

~/anaconda3/lib/python3.7/site-packages/tensorflow_probability/python/distributions/transformed_distribution.py in _call_sample_n(self, sample_shape, seed, name, **kwargs)
    418       # returned result.
    419       y = self.bijector.forward(x, **bijector_kwargs)
--> 420       y = self._set_sample_static_shape(y, sample_shape)
    421 
    422       return y

~/anaconda3/lib/python3.7/site-packages/tensorflow_probability/python/distributions/distribution.py in _set_sample_static_shape(self, x, sample_shape)
   1420         shape = tf.TensorShape([None]*sample_ndims).concatenate(
   1421             self.batch_shape).concatenate([None]*event_ndims)
-> 1422         tensorshape_util.set_shape(x, shape)
   1423 
   1424     return x

~/anaconda3/lib/python3.7/site-packages/tensorflow_probability/python/internal/tensorshape_util.py in set_shape(tensor, shape)
    317   """
    318   if hasattr(tensor, 'set_shape'):
--> 319     tensor.set_shape(shape)
    320 
    321 

~/anaconda3/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py in set_shape(self, shape)
   1086       raise ValueError(
   1087           "Tensor's shape %s is not compatible with supplied shape %s" %
-> 1088           (self.shape, shape))
   1089 
   1090   # Methods not supported / implemented for Eager Tensors.

ValueError: Tensor's shape (2,) is not compatible with supplied shape ()
```

Sayam753:
Hi @sshkhr Looking at the traceback, it seems that the batch_shape for `Uniform` distribution does not broadcast with the bijector. You can explicitly do this.
```python
>>> dist = tfd.TransformedDistribution(distribution=tfd.Uniform(low=[0, 0], high=[1, 1]), 
...                                    bijector=tfb.Invert(tfb.WeibullCDF(scale=[1, 1], concentration=[1, 1])), 
...                                    name="Weibull")
>>> print(dist.sample())
tf.Tensor([0.2631347  0.30459547], shape=(2,), dtype=float32)
```

sshkhr:
Thanks a ton, @Sayam753 for the heads up. I have made some changes and this looks okay from my side. It'd be great if someone could review the PR

pymc4/distributions/continuous.py
```diff
@@ -1232,54 +1233,69 @@ def upper_limit(self):
...
+
+    ========  ====================================================
+    Support   :math:`x \in [0, \infty)`
+    Mean      :math:`\beta \Gamma(1 + \frac{1}{\alpha})`
+    Variance  :math:`\beta^2 \Gamma(1 + \frac{2}{\alpha} - \mu^2)`
+    ========  ====================================================
+
+    Parameters
+    ----------
+    low : float|tensor
```
lucianopaz: `low` and `high` should not be parameters of the Weibull distribution.

pymc4/distributions/continuous.py
```diff
@@ -1232,54 +1233,69 @@ def upper_limit(self):
...
+        Upper limit of underlying uniform distribution (Use 1 for standard Weibull distribution).
+    concentration : float|tensor
+        Shape parameter (concentration > 0).
+    scale : float|tensor
+        Scale parameter (scale > 0).
+    """
+
+    def __init__(self, name, low, high, concentration, scale, **kwargs):
+        super().__init__(
+            name, low=low, high=high, concentration=concentration, scale=scale, **kwargs
```
lucianopaz: You'll have to remove `low` and `high` from the functions signature.

pymc4/distributions/continuous.py
```diff
@@ -1232,54 +1233,69 @@ def upper_limit(self):
...
+        )
+
+    @staticmethod
+    def _init_distribution(conditions):
+
+        low, high, concentration, scale = (
```
lucianopaz: The user must not pick either `low` or `high`. You must set them as `tf.zeros(shape)` and `tf.ones(shape)` respectively. The `shape` must be equal to the shape to which `concentration` and `scale` will broadcast to.

tests/test_distributions.py
```diff
@@ -260,10 +260,15 @@
...
         },
     },
-    # "Weibull": {
-    #     "scalar_parameters": {"beta": 0.1, "alpha": 1.0},
-    #     "multidim_parameters": {"beta": 0.1, "alpha": 1.0},
-    # },
+    "Weibull": {
+        "scalar_parameters": {"low": 0.0, "high": 1.0, "concentration": 1.0, "scale": 1.0},
+        "multidim_parameters": {
+            "low": np.array([0.0, 0.0], dtype="float32"),
```
lucianopaz: As above, you have to drop the low and high parameters

sshkhr:
Thanks a lot, @lucianopaz and @Sayam753 for the feedback and help. I have made the requested changes. Please let me know if this looks fine.

On another note, I am looking to apply for the VI project with PyMC4 for GSoC. I know you're one of the potential mentors Luciano so I was wondering where would be a good place to discuss about it. I also happen to know @ferrine from DeepBayes in Moscow last year. I dunno if Max remembers me though :sweat_smile:

pymc4/distributions/continuous.py
sshkhr: Removed :+1:

pymc4/distributions/continuous.py
sshkhr: Removed :+1:

pymc4/distributions/continuous.py
sshkhr: I have used `tf.zeros_like()` and `tf.ones_like()` and used the shape of `scale`

tests/test_distributions.py
sshkhr: Done :+1:

ferrine:
I do remember we had lunch)

pymc4/distributions/continuous.py
lucianopaz: I think that this isn't enough. Image if the user speficies a scalar `scale` but a vector `concentration`. This will run into exactly the same issue you were seeing before when you passed scalar values to `low` and `high`. What you have to do, is to look at the `shape` that would come out of broadcasting `scale` and `concentration`, and use that `shape` with `tf.zeros` and `tf.ones`.

lucianopaz:
> Thanks a lot, @lucianopaz and @Sayam753 for the feedback and help. I have made the requested changes. Please let me know if this looks fine.
> 
> On another note, I am looking to apply for the VI project with PyMC4 for GSoC. I know you're one of the potential mentors Luciano so I was wondering where would be a good place to discuss about it. I also happen to know @ferrine from DeepBayes in Moscow last year. I dunno if Max remembers me though 😅

I'm not sure if I will be available to mentor during the summer. Some other people have already said that they were interested on working in VI for pymc4. Would you like to consider another pymc4 project listed [here](https://github.com/pymc-devs/pymc3/wiki/GSoC-2020-projects)? I'm not really sure how the application goes but [quoting @twiecki](https://discourse.pymc.io/t/regarding-variational-inference-project-for-gsoc-2020/4525/3), you'll need to write a short proposal that we can review, and then formally apply to GSoC.

sshkhr:
@ferrine That's awesome to hear

@lucianopaz No worries. I am taking a class on MCMC this semester so I am interested in the first project as well. I would like to know the scale of what is expected in the project because afaik PyMC4 only implements NUTS while TFP has quite a big mcmc suite. #229 seems to be a WIP regarding more samplers so maybe I will start by looking at that and also start a discussion with the mentors for the samplers project on Discourse :+1: 

Also, could one of the maintainers please review this PR when you get a chance. Thanks

lucianopaz:
> Also, could one of the maintainers please review this PR when you get a chance. Thanks

@sshkhr, I had left a comment saying that the `tf.zeros_like` was not enough. I'll copy it down here:

> I think that this isn't enough. Image if the user speficies a scalar `scale` but a vector `concentration`. This will run into exactly the same issue you were seeing before when you passed scalar values to `low` and `high`. What you have to do, is to look at the `shape` that would come out of broadcasting `scale` and `concentration`, and use that `shape` with `tf.zeros` and `tf.ones`.

pymc4/distributions/continuous.py
sshkhr: shape = tf.broadcast_static_shape(tf.constant(concentration).shape,tf.constant(scale).shape)

low = tf.zeros(shape)
high = tf.ones(shape)

Does this look like a good workaround?

lucianopaz: No, don't use `tf.constant`, use `tf.convert_to_tensor` instead. Then, I'm not really sure if these will be static shapes, maybe @junpenglao could comment on this?

sshkhr: There's also `tf.broadcast_dynamic_shape` https://www.tensorflow.org/api_docs/python/tf/broadcast_dynamic_shape

lucianopaz: I tried to dig into how tfp does its broadcasting and I found that they [defined some helper functions](https://github.com/tensorflow/probability/blob/a914215150d14f2a3a4e3e5f554435e36a847e2a/tensorflow_probability/python/internal/distribution_util.py#L971-L1052) to broadcast tensors or to simply get their ranks and shapes, preferring the static shapes over the dynamic ones. You could import and use those.

PR #227, opened 2020-02-17
Add meta executor

### LINKED ISSUE #167: Over reliance on evaluate_model

lucianopaz:
This PR closes #167.

It adds the `MetaSamplingExecutor` and `MetaPosteriorPredictiveSamplingExecutor`, that work exactly the same as the previous executors but use `distribution.test_sample` instead of `distribution.sample`. This will allow us to use `Flat` and `HalfFlat` distributions in our models and still be able to `pm.sample` and `pm.sample_posterior_predictive` from them.

### DIFF
--- a/pymc4/distributions/distribution.py
+++ b/pymc4/distributions/distribution.py
@@ -123,6 +123,24 @@ class Distribution(Model):
         """
         return self.sample(sample_shape, seed).numpy()
 
+    def test_sample(self, sample_shape=(), seed=None):
+        """Get the test value using a function signature similar to meth:`~.sample`
+        
+        Parameters
+        ----------
+        sample_shape : tuple
+            sample shape
+        seed : int | None
+            ignored. Is only present to match the signature of meth:`~.sample`
+        
+        Returns
+        -------
+        The distribution's ``test_value`` broadcasted to
+        ``sample_shape + self.batch_shape + self.event_shape``
+        """
+        sample_shape = tf.TensorShape(sample_shape)
+        return tf.broadcast_to(self.test_value, sample_shape + self.batch_shape + self.event_shape)
+
     def log_prob(self, value):
         """Return log probability as tensor."""
         return self._distribution.log_prob(value)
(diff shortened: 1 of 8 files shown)


tests/test_executor.py
```diff
@@ -157,7 +157,7 @@ def nested_model(cond):
 
     @pm.model
     def outer_model():
-        cond = yield pm.HalfNormal("cond", 1)
+        cond = yield pm.HalfNormal("cond", 1, conditionally_independent=True)
```
ColCarroll: what happens without this condition?

pymc4/distributions/distribution.py
```diff
@@ -123,6 +123,24 @@ def sample_numpy(self, sample_shape=(), seed=None):
         """
         return self.sample(sample_shape, seed).numpy()
 
+    def test_sample(self, sample_shape=(), seed=None):
```
ColCarroll: can this have a more descriptive name? I assumed this was testing sample...
```suggestion
    def get_test_value(self, sample_shape=(), seed=None):
```

tests/test_executor.py
lucianopaz: If this wasn't true, then the `sample_shape` that gets passed to the executor is ignored. I wanted to test how the executor works when it uses the supplied `sample_shape` (when `conditionally_independent=True`) and when it is ignored (all other distributions).

pymc4/distributions/distribution.py
lucianopaz: I'll change it today

Issue #187, opened 2019-12-16
Support continuous and discrete variables


ColCarroll:
Related to #171, exposing more TFP step methods through the API, and automatically assigning a reasonable sampler to each variable. PyMC3 uses a `competence` attribute to assign this.

Issue #171, opened 2019-12-03
Only one sampler is available


rrkarim:
I don't see any issues to support more samplers. Can anyone be assigned to the task or somebody already in progress of completing it?

lucianopaz:
@rrkarim, it would be great to add more samplers! We should think of a way to easily get the other tfp transition kernels

rrkarim:
Yeah, I can think of this, will post the solution here.

PR #229, opened 2020-02-19
[WIP] Add discrete/continuous support+more samplers

rrkarim:
addr #187 #171
PR should include:
- [x] Support for more samplers (the design should be discussed)
- [x] Support for compound step sampling (maybe it is hard one to implement because of over-reliance on batching in tfp)
- [x] Support for discrete var sampling
- [ ] Support `auto_assign_method` based on a set of certain criterias (like in `pymc3`)
- [ ] More reach trace for compound step
- [ ] More tests (they suck for now)

### Current user design of compound step for user looks like this:
```python
# the easiest one with automatic method assignment
# for each free var
model = ... # we define the model
pm.sample(model(), sampler_type="compound")
# user defines the methods for selected vars in model
from pymc4.mcmc.samplers import NUTS, RandomWalkM
pm.sample(model(), 
    sampler_type="compound", 
    sampler_methods=[
        ("var", RandomWalkM), 
        ("mean", NUTS)
    ], 
    xla=True)
```
There is currently support for categorical distribution mcmc sampling with custom state function, implemented in `pymc4/distributions/state_functions.py`, but more distributions should be tested and supported.
- [x] Categorical
- [ ] **should be added as its tested and supported


rrkarim:
There are some problem with sampling discrete distributions with `xla` enabled. I will look at the issue closer.

pymc4/distributions/distribution.py
```diff
@@ -2,8 +2,10 @@
 import copy
 from typing import Optional, Union, Any
 
+from absl import logging
```
fonnesbeck: Do we need/want this as a dependency?

rrkarim: Isn't the `tf.logging` replaced with `absl-py` in `TF 2.0`?

pymc4/distributions/distribution.py
```diff
@@ -195,6 +197,19 @@ def batch_shape(self):
     def event_shape(self):
         return self._distribution.event_shape
 
+    @property
+    def grad_support(self, state=None):
+        if state is None:
+            state = self.sample()
+        try:
+            logging.set_verbosity(logging.ERROR)
```
rrkarim: @fonnesbeck It hides additional tf grad errors here.

rrkarim:
@junpenglao can you please provide some feedback on the issues #233, I've implemented the sampling of the models with categorical distribution (with support of `new_state_fn`) but the implementation still relies on `CompounStep`. There is a brief notebook that will test sampling.

rrkarim:
Sorry for some notebooks edited, tests do not pass without it. There are also some issues with notebook testing, radom example doesn't work on my local machine.

rrkarim:
So basically, I had two ways of designing sampler methods. This PR implementation is similar to `pymc3`, which for me is more clear from user perspective. Another idea was to have `MCMCSampler` class with all the ops like `sample, run_chains etc.` and sampler subclasses that defined the parameters like `kernel`, `adaptive_kernel`, `trace_fn` and etc. Then `_assign_default_methods` method should be defined in `MCMCSampler` class (it is defined now in `CompoundStep`) so that user doesn't pass `sampler_type` argument, but the samplers are assigned automatically to each free var according to certain criterias. Basically, all the sampler methods are now the type of compound step. What I don't like about this design choice is how the parametrization for each sampler method is handled. S

fonnesbeck:
I assume `Unititled1.ipynb` is not supposed to be there.

PR #237, opened 2020-03-26
Add Truncated Normal distribution

ritesh99rakesh:
I was looking to contribute to PyMC4 and based on #44 I thought I would add the discrete Truncated Normal distribution for a start. When I looked at the source it seems to me that the Truncated Normal distribution is not available either.

This is my first contribution to PyMC so please let me know if there are any particular tests you'd like me to add and if something needs to be fixed in the PR.

### DIFF
--- a/pymc4/distributions/continuous.py
+++ b/pymc4/distributions/continuous.py
@@ -1119,6 +1119,7 @@ class Triangular(BoundedContinuousDistribution):
     def upper_limit(self):
         return self.conditions["high"]
 
+
 class TruncatedNormal(ContinuousDistribution):
     r"""Univariate truncated normal random variable.
 
@@ -1209,12 +1210,13 @@ class TruncatedNormal(ContinuousDistribution):
         loc, scale = conditions["loc"], conditions["scale"]
         low, high = conditions["low"], conditions["high"]
         return tfd.TruncatedNormal(loc=loc, scale=scale, low=low, high=high, validate_args=True)
-    
+
     def lower_limit(self):
         return self.conditions["low"]
 
     def upper_limit(self):
         return self.conditions["high"]
+
 
 class Uniform(BoundedContinuousDistribution):
     r"""Continuous uniform random variable.
--- a/pymc4/distributions/continuous.py
+++ b/pymc4/distributions/continuous.py
@@ -34,6 +34,7 @@ __all__ = [
     "Pareto",
     "StudentT",
     "Triangular",
+    "TruncatedNormal",
     "Uniform",
     "VonMises",
     "HalfStudentT",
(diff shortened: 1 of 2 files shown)


ericmjl:
I’ll chime in from my lunchtime - first off, thanks for the PR! :tada:

If you study the tests that are present, I think we try to make sure that we can at least sample from the distribution. Perhaps give that a shot?

ritesh99rakesh:
Hi @ericmjl! Could you please let me know why tests are failing. I tried but could not understand what is wrong. Thanks!

pymc4/distributions/continuous.py
```diff
@@ -1119,6 +1120,104 @@ def upper_limit(self):
...
+    --------
+    .. code-block:: python
+        @pm.model
+        def model():
+            x = pm.TruncatedNormal('x', loc=0, scale=10, low=-8, high=5)
+    """
+
+    def __init__(self, name, loc, scale, **kwargs):
+        super().__init__(name, loc=loc, scale=scale, low=low, high=high, **kwargs)
+
```
Sayam753: Hi @ritesh99rakesh 
One possible reason that tests are failing because `low` and `high` arguments are not specified in `__init__` method.

tests/test_distributions.py
```diff
@@ -262,6 +262,15 @@
             "peak": np.array([0.5, 0.5], dtype="float32"),
         },
     },
+    "TruncatedNormal": {
+        "scalar_parameters": {"loc": 0.0, "scale": 1.0, "low": 0.0, "high": 2.0},
+        "multidim_parameters": {
+            "loc": np.array([0.0, 0.0], dtype="float32"),
+            "scale": np.array([1.0, 1.0], dtype="float32"),
+            "low": np.array([0.0, 0.0]),
+            "high": np.array([2.0, 2.0]),
```
Sayam753: Another thing to look is `common_dtype` method [here](https://github.com/tensorflow/probability/blob/v0.9.0/tensorflow_probability/python/distributions/truncated_normal.py#L143). You may run into compatibility issues between `np.float32` and `np.float64`.

ritesh99rakesh:
Thank you for comments. I also wanted to know how can I run the tests locally?

Sayam753:
I have `pytest` installed in my virtualenv. 
- For running all tests - `pytest -v tests`
- For running a specific test, 

Type `pytest --collect-only`  in the base directory for `pymc4` to capture test function signature and run the test. 
Example - 
```bash
pytest -v tests/test_distributions.py::test_rvs_logp_and_forward_sample[scalar_parameters-AR]
```
I am not sure if there is any other way to run tests locally!

lucianopaz:
You can also run the tests locally by running `make test` in the project's root directory

PR #238, opened 2020-03-26
Move test_sampling fixtures to conftest file

### LINKED ISSUE #194: Centralize fixtures
As pymc4 grows, the test suite gets bigger and we add more fixtures to the test scripts. I think that it would be good practice to start to move the fixtures out of the test scripts and into either a central `conf.py` or separate modules that get imported into a `conf.py`. Some fixtures are awfully similar across test files and we may be able to fuse them into a single coherent fixture.

techytushar:
Fixes #194
Moved all the fixture functions in `test_sampling.py` to `conftest.py`.
@lucianopaz @twiecki Is this how you want this to be implemented or is there any other way. I have currently done it for one file, I can do it for rest of them too.


twiecki:
I think that looks great!

lucianopaz:
Thanks for this contribution, @techytushar, but all the rest of the test scripts have fixtures too. It would be great if they could all be moved to conftest.py and maybe refactored a bit to reduce duplicate code. I'll reopen #194.

PR #239, opened 2020-03-27
Validate args

### LINKED ISSUE #236: Distributions do not validate argument
Many distributions do not validate arguments:
```python
>>> import pymc4
>>> x = pymc4.distributions.Normal('x', loc=0, scale=-2)
>>> x.sample()
<tf.Tensor: shape=(), dtype=float32, numpy=4.025787>
>>> x = pymc4.distributions.Uniform('x', low=0, high=-2)
>>> x.sample()
<tf.Tensor: shape=(), dtype=float32, numpy=-0.24736929>
```

One way to solve would be set `validate_args=True` in TensorFlow distribution calls:
```python
@staticmethod
def _init_distribution(conditions):
    loc, scale = conditions["loc"], conditions["scale"]
    return tfd.Normal(loc=loc, scale=scale, validate_args=True)
```
which would give the following error:
```python
 = pymc4.distributions.Normal('x', loc=0, scale=-2)
Traceback (most recent call last):
...
Condition x > 0 did not hold element-wise:
x (shape=() dtype=float32) = 
-2.0
```

If the community thinks that this is a valid issue then I can work on it.

ritesh99rakesh:
Fixes #236 

I have set `validate_args=True` in TensorFlow distribution calls:
```python
@staticmethod
def _init_distribution(conditions):
    loc, scale = conditions["loc"], conditions["scale"]
    return tfd.Normal(loc=loc, scale=scale, validate_args=True)
```
Add `validate_args=True` for each distribution types:
- [x] Continuous
- [ ] Discrete
- [ ] Half Student
- [ ] Others

### DIFF
--- a/pymc4/distributions/continuous.py
+++ b/pymc4/distributions/continuous.py
@@ -104,7 +104,7 @@ class Normal(ContinuousDistribution):
     @staticmethod
     def _init_distribution(conditions):
         loc, scale = conditions["loc"], conditions["scale"]
-        return tfd.Normal(loc=loc, scale=scale)
+        return tfd.Normal(loc=loc, scale=scale, validate_args=True)
 
 
 class HalfNormal(PositiveContinuousDistribution):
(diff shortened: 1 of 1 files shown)


lucianopaz:
Thanks for the initiative, @ritesh99rakesh, but I disagree with this design. Validate args should be optional, and should follow a common pattern that would be defined in the distribution base class. The tfd distribution initialization will have to be modified though, but the design should enable us to easily add in future common base class parameters, like maybe dtype or allow Nan stats. The current state of this PR doesn't allow that. Would you like to think of an alternative?

lucianopaz:
Closed in favor of #281

PR #242, opened 2020-05-07
Fixed typos in utils.py and flow/executor.py

Sayam753:
Summary of changes - 
- Added `pm.distributions.Deterministic` to `validate_return_object` function.
- Changed the word `fn` [here](https://github.com/pymc-devs/pymc4/blob/master/pymc4/utils.py#L29) to be more specific as it is one of the arguments name.


fonnesbeck:
Thanks!

PR #243, opened 2020-05-08
Update scopes.py docstring

ferrine:
@Sayam753 suggested that there is misleading documentation and what is said in doc is not actually done. I agree with the point and modify the docstring to make it more clear. If any suggestions to improve this case please comment.


Sayam753:
And one more typo in line [43](https://github.com/pymc-devs/pymc4/blob/master/pymc4/scopes.py#L43), `cls.contexts` should be `cls.context`.

PR #246, opened 2020-05-15
Add exp transform

ritesh99rakesh:
I have added `Exp` transform under `pymc4/distributions/transforms.py`.

### DIFF
--- a/pymc4/distributions/transforms.py
+++ b/pymc4/distributions/transforms.py
@@ -160,13 +160,13 @@ class Exp(Transform):
 
     def __init__(self):
         self._transform = tfb.Exp()
-    
+
     def forward(self, x):
         return self._transform.inverse(x)
-    
+
     def inverse(self, z):
         return self._transform.forward(z)
-    
+
     def forward_log_det_jacobian(self, x):
         return self._transform.inverse_log_det_jacobian(x, self._transform.inverse_min_event_ndims)
 
--- a/pymc4/distributions/transforms.py
+++ b/pymc4/distributions/transforms.py
@@ -3,7 +3,7 @@ from typing import Optional
 
 from tensorflow_probability import bijectors as tfb
 
-__all__ = ["Log", "Sigmoid"]
+__all__ = ["Log", "Sigmoid", "Exp"]
 
 
 class JacobianPreference(enum.Enum):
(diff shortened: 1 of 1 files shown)


ritesh99rakesh:
If the current code looks good then I will add more transforms.
I have just started to learn transform and I find it a good opportunity to work on pymc4 library.

Please review this PR.

ritesh99rakesh:
All the checks are passing now. Can someone please review it.

lucianopaz:
@ritesh99rakesh, first of all thanks for your eagerness to contribute to pymc4. Every bit of help counts and I'm sorry that I was not able to review your PR sooner.

About what you wrote, I'm sorry to tell you that its exactly the same as our `Log` transform. The `Exp` transformed could be implemented as `inverse(Log)`, but I'll try to explain why this is not actually desirable.

We have had some internal discussions for some time now about the name of the `transforms` module. We have arrived to the conclusion that calling this module and the contents therein _transforms_ is misleading. They **are not** operations that transform samples drawn from the distribution. They **are** operations that grab values that can be any real number (their domain is the entire real line) and in the cases of multidimensional distributions they can be vectors in $R^{n}$, and transform these values into the distribution's support. Furthermore, these operations are **only used** while performing inference with MCMC or Variational inference, but they are never used while generating samples in what we call _forward passes_ (sample from prior and posterior predictive distributions).

To try to make this more explicit I'll write down a bit of math on how the `Log` transform is applied. Lets suppose we have a `distribution` and the samples that we draw from it are `x`. If we apply the `Log` transform to this distribution, the operations we end up with are:

$$
x \sim distribution(\theta)
y = \log(x)
$$

where $\theta$ represent any distribution parameters. When we perform inference with MCMC methods, the step methods propose values of $y$, and then evaluate the logarithm of the probability density function as:

$$
distribution.log_prob(\exp(y), \theta)
$$

Why do we even want to make proposals for values of `y` and not work with `x` directly? Because if we work with `y`, we don't have to worry about imposing any constraints to its value to ensure the proposals are supported by the distribution, they will work as they are thanks to the inverse of the transform operation. As you can see, the transform is actually a bit odd because the inverse of our `Log` transform is what we use in the step methods for MCMC, and then it is never used in forward samples. So if we define an `Exp` transform, we would use the `log` of the proposed values, that will run into trouble when the proposals are negative.

ritesh99rakesh:
Thank you for explanation @lucianopaz 
I will close my PR. Also, I am quite interested in contributing towards `pymc4` and therefore it would be generous of you if you could guide me of a possible area of contribution.

lucianopaz:
@ritesh99rakesh, about contributing, it would be really great if you could continue #239, but making it more general than just `validate_args`.

PR #252, opened 2020-05-28
More distributions

dirmeier:
Hello,

this PR adds some TFP distributions (LKJ Cholesky, multivariate normal Cholesky, ordered logit) I frequently need and fixes a typo in the documentation, in case you want to merge it.

I switched from Stan to TFP a while ago and would love to contribute more code/documentation/notebooks/.. to PyMC4, if you would be interested.

Cheers,
Simon

### DIFF
--- a/pymc4/distributions/discrete.py
+++ b/pymc4/distributions/discrete.py
@@ -14,6 +14,7 @@ __all__ = [
     "Categorical",
     "Geometric",
     "NegativeBinomial",
+    "OrderedLogistic",
     "Poisson",
     "Zipf",
 ]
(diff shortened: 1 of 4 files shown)


pymc4/distributions/discrete.py
```diff
@@ -709,3 +710,46 @@ def __init__(self, name, power, **kwargs):
...
+    """
+
+    def __init__(self, name, loc, cutpoints, **kwargs):
+        super().__init__(name, loc=loc, cutpoints=cutpoints, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions):
+        cutpoints = tf.convert_to_tensor(conditions["cutpoints"])
+        loc = conditions["loc"]
+        return tfd.OrderedLogistic(cutpoints=cutpoints, loc=loc, validate_args=True)
```
eigenfoo, line 749: @junpenglao do we have an opinion on whether the distributions should `validate_args`? It seems safest to me, but we should probably be consistent. In any case, we can handle this in a follow-up PR.

lucianopaz: My opinion is that this would be optional but we have change the init_distribution signature to make it easy to pass other tfd options, like dtype.
For example, see the conversation on #239

pymc4/distributions/discrete.py
```diff
@@ -709,3 +710,46 @@ def __init__(self, name, power, **kwargs):
...
+
+    def lower_limit(self):
+        return 0
+
+    def upper_limit(self):
+        return len(self.conditions["cutpoints"])
```
lucianopaz, line 755: We shouldn't use `self.conditions`. We should look at the resulting `_distribution` cutpoints attribute instead.

dirmeier: Right, sorry! The other distributions use `self.conditions`, too. Should I change them as well?

lucianopaz: That would be great. But it would be best to do so in a separate PR.

dirmeier: > This is great. Could you also add them to our test suite in test_distributions.py?

Yeah, for sure. The last commit adds test cases for the three distributions and worked in the other changes you and @eigenfoo requested. Thanks!

dirmeier: All right.

dirmeier:
The last commit adds test cases for the three distributions and worked in the other changes you and @eigenfoo requested. Thanks!

eigenfoo:
Thanks for the PR, and congratulations on the contributions!

Follow up PRs:

- [ ] Use `_distribution` instead of `self.conditions`
- [ ] Do not pass `validate_args=True`: see #236

PR #235, opened 2020-03-01
[MRG] ENH: Add Basic Gaussian Process Interface

tirthasheshpatel:
According to the discussion on [this discourse thread](https://discourse.pymc.io/t/adding-gaussian-processes-in-pymc4-this-summer/4538), I have created a simple GP interface for PyMC4.

- [x] add base classes for covariance functions, mean functions, gp.
- [x] add a working ``LatentGP`` model for univariate and multivariate data.
- [x] add tests for ``LatentGP``
- [x] add tests for mean functions and covariance functions
- [x] add tests for a combination of mean and covariance functions
- [x] ~add an interface for combining GP models~
- [x] ~error handling for invalid shapes (maybe?)~
- [x] add a notebook with pretty plots.
- [x] add ``__radd__`` and ``__rmul__`` methods for combination of covariance functions with non-covariance functions.
- [x] add tests for ``__radd__`` and ``__rmul__`` methods.
- [x] add typing in all files

Thanks, @fonnesbeck for your suggestions and help! I think this a better place for discussion for further development.
Comments and criticism are welcome!

PR #269, opened 2020-05-29
MAINT: add reparameterize to gp and fix mvnormal

tirthasheshpatel:
I changed `MvNormal` to use cholesky component of full covariance matrix in #235 but #252 introduced `MvNormalCholesky` which is more appropriate to use instead.

Also, parameters like `covariance_matrix` are very explicit by name and shorthand name like `cov` is more convenient and obvious enough. I have tried to address it in this pr but if there is a reason behind such names, I will revert those changes.

I have also started fixing docs of `MvNormal` and `MvNormalCholesky` and stuff in the gp module.

### DIFF
--- a/pymc4/distributions/multivariate.py
+++ b/pymc4/distributions/multivariate.py
@@ -177,39 +177,31 @@ class MvNormal(ContinuousDistribution):
 
     Parameters
     ----------
-    loc : array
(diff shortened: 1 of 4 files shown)


tirthasheshpatel:
Closed in favour of #296

PR #276, opened 2020-06-02
support generalized normal distribution version 1

### LINKED ISSUE #275: Implement GeneralizedNormal distribution
GeneralizedNormal distribution is added to tfp recently. I would like to add the corresponding class to PyMC4 as my first contribution. Do you think this would be a useful addition?

There are also some other new distributions in tfp (Bates, Moyal and JohnsonSU). I can also open issues for these distributions if you are interested.

kori73:
Closes #275 
Add GeneralizedNormal class to distributions.
Add test parameters to test_distributions.

I am not sure if the name "GeneralizedNormal" is okay since pymc-devs are expecting SkewNormal implementation in tfp as well.

### DIFF
--- a/pymc4/distributions/continuous.py
+++ b/pymc4/distributions/continuous.py
@@ -21,6 +21,7 @@ __all__ = [
     "Chi2",
     "Exponential",
     "Gamma",
+    "GeneralizedNormal",
     "Gumbel",
     "HalfCauchy",
     "HalfNormal",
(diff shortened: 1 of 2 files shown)


lucianopaz:
This is great @kori73! Thanks for contributing and you are very welcome to try to add some more distributions.

pymc4/distributions/continuous.py
```diff
@@ -110,6 +111,67 @@ def _init_distribution(conditions):
...
+       f(x \mid \mu, \alpha, \beta) = 
+           \frac{\beta}{2 \Gamma(1/\beta)}
+           \exp(-(|x - \mu| /\alpha)^\beta)
+
+    .. plot::
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import scipy.stats as st
+        plt.style.use('seaborn-darkgrid')
```
aloctavodia, line 130: Hi @kori73 , thanks for your contribution! PyMC4 does not longer requieres seaborn. This should be plt.style.use('arviz-darkgrid'). If you are planning to keep adding distributions you may change this in your next PR.

kori73: Sure, I will do that in my next PR.

PR #278, opened 2020-06-04
Add support for Moyal distribution

### LINKED ISSUE #277: Add support for Moyal distribution
Add support for Moyal distribution (recently added to tfp).

kori73:
add Moyal class to continuous.py.
add its test dictionary.

use arviz-darkgrid style in GeneralizedNormal.
Closes #277

### DIFF
--- a/pymc4/distributions/continuous.py
+++ b/pymc4/distributions/continuous.py
@@ -992,7 +992,7 @@ class Moyal(ContinuousDistribution):
         import numpy as np
         import scipy.stats as st
         plt.style.use('arviz-darkgrid')
-        x = np.linspace(-5, 5, 1000)
+        x = np.linspace(-5, 10, 1000)
         locs = [0., 0., 0., -2.]
         scales = [0.5, 0.6, 1., 1.]
         for loc, scale in zip(locs, scales):
--- a/pymc4/distributions/continuous.py
+++ b/pymc4/distributions/continuous.py
@@ -128,7 +128,7 @@ class GeneralizedNormal(ContinuousDistribution):
         import matplotlib.pyplot as plt
         import numpy as np
         import scipy.stats as st
-        plt.style.use('seaborn-darkgrid')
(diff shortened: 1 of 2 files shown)


pymc4/distributions/continuous.py
```diff
@@ -974,6 +975,64 @@ def _init_distribution(conditions):
...
+    Parameters
+    ----------
+    loc : float
+        Location parameter.
+    scale : float
+        Scale parameter.
+
+    Examples
+    --------
+    .. code-block:: python
```
tirthasheshpatel, line 1021: can you please do something like:
```python
Examples
----------
>>> @pm.model
... def model():
...     x = pm.Moyal('x', loc=0, scale=10)
```

It would be helpful while compiling docs because ``.. code-block::`` directive doesn't work in the Examples section and doc tests fail. Although it is fine as of now (you can make another or for that, if you want to) because all the distributions have this doc style which needs to be fixed.

kori73: Sure, I have changed the example section as you suggested.

pymc4/distributions/continuous.py
```diff
@@ -974,6 +975,63 @@ def _init_distribution(conditions):
...
+           \frac{1}{\sqrt{2\pi}\sigma}
+           \exp\left(-\frac{1}{2}\left[\frac{x-\mu}{\sigma}+\exp\left(-\frac{x-\mu}{\sigma}\right)\right]\right)
+
+    .. plot::
+
+        import matplotlib.pyplot as plt
+        import numpy as np
+        import scipy.stats as st
+        plt.style.use('arviz-darkgrid')
+        x = np.linspace(-5, 5, 1000)
```
aloctavodia, line 995: nitpicking: Could you use `x = np.linspace(-5, 8, 1000)` or `x = np.linspace(-5, 10, 1000)`?

kori73: done

PR #284, opened 2020-06-16
Implement mixture distribution

dirmeier:
Hello all,

I've started implementing mixture distributions, in case you're interested. However, there are still some issues:

- the TFP mixture uses latent, discrete indicators and I don't see how one can easily implement marginal mixtures using it,
- the posterior intervals of some parameters are different when I use different implementations and I don't yet understand why that is. More specifically, imagine if I implement the mixture class and a model like this:

```python
class Mixture(Distribution):
    def __init__(self, name, p, distributions, **kwargs):
        super().__init__(name, p=p, distributions=distributions, **kwargs)

    @staticmethod
    def _init_distribution(conditions, **kwargs):
        p, d = conditions["p"], conditions["distributions"]
        distributions = [el._distribution for el in d]
        return tfd.Mixture(cat=tfd.Categorical(probs=p),
                           components=distributions,
                           **kwargs)
@pm.model
def mixture(dat):
    p = [0.5, 0.5]
    m = yield pm.Normal("means", loc=[0.0, 0.0], scale=1.0)
    comp1 = pm.Normal("comp1", m[0], scale=1.0)
    comp2 = pm.Normal("comp2", m[1], scale=1.0)
    obs = yield Mixture("mix", p=p, distributions=[comp1, comp2], observed=dat)
    return obs
```

In this case, I use the underlying TFP distribution of a PyMC distribution and just pass this to `tfd.Mixture`. Hence, I can't use a `yield` statement in the model block for the components (since I need access to the distributions).

Alternatively, I can implement the mixture class like this:

```python
class MixtureManual(Distribution):
    def __init__(self, name, p, distributions, **kwargs):
        super().__init__(name, p=p, distributions=distributions, **kwargs)

    @staticmethod
    def _init_distribution(conditions, **kwargs):
        p, distributions = conditions["p"], conditions["distributions"]
        return tfd.Mixture(cat=tfd.Categorical(probs=p),
                           components=[
                               tfd.Normal(loc=distributions[0], scale=1.0),
                               tfd.Normal(loc=distributions[1], scale=1.0)],
                           **kwargs)

@pm.model
def mixture_manual(dat):
    p = [0.5, 0.5]
    m = yield pm.Normal("means", loc=[0.0, 0.0], scale=1.0)
    comp1 = yield pm.Normal("comp1", m[0], scale=1.0)
    comp2 = yield pm.Normal("comp2", m[1], scale=1.0)
    obs = yield MixtureManual("mix", p=p, distributions=[comp1, comp2], observed=dat)
    return obs
```

In this case, I construct the TFP distributions in the mixture class on the fly.  This version works on `tensors` (*with* `yield` statements in the model block). 

Interestingly, the inferences are not the same. When I create data using a two-component Gaussian mixture with means at `-5` and `5` I get the following posteriors. 

inf_mix

While the means are correctly inferred in the left subfigure, the posteriors on the right 'look more correct'. Am I doing sth wrong here? Here's a [Colab](https://colab.research.google.com/drive/1oyB4lFaqzltgFS3Yz-vDmLQ4AYQxBpfL?usp=sharing) that shows this.

Help would be much appreciated. 

Cheers,
Simon

### DIFF
--- a/pymc4/distributions/mixture.py
+++ b/pymc4/distributions/mixture.py
@@ -1,7 +1,15 @@
+"""PyMC4 Distribution of a random variable consisting of a mixture of other
+distributions.
+
+Wraps tfd.Mixture as pm.Mixture
+"""
+import tensorflow as tf
+from tensorflow_probability import distributions as tfd
+
+import pymc4 as pm
 from pymc4.distributions.distribution import Distribution
 
 
-# TODO: Implement this
 class Mixture(Distribution):
     r"""
     Mixture random variable.
@@ -19,7 +27,7 @@ class Mixture(Distribution):
     ----------
     p : array of floats
         p >= 0 and p <= 1
-        the mixture weights, in the form of probabilities
+        the mixture weights, in the form of probabilities, must sum to one.
     distributions : multidimensional PyMC4 distribution (e.g. `pm.Poisson(...)`)
         or iterable of one-dimensional PyMC4 distributions the
         component distributions :math:`f_1, \ldots, f_n`
@@ -27,3 +35,16 @@ class Mixture(Distribution):
 
     def __init__(self, name, p, distributions, **kwargs):
         super().__init__(name, p=p, distributions=distributions, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions):
+        p, distributions = conditions["p"], conditions["distributions"]
+        if isinstance(p, pm.Categorical):
+            cat = p._distribution
+        else:
+            cat = tfd.Categorical(probs=p)
+        if isinstance(distributions, list):
+            distributions = [d._distribution for d in distributions]
+        elif isinstance(distributions, Distribution):
+            return
+        return tfd.Mixture(cat=cat, components=distributions)
--- a/pymc4/distributions/mixture.py
+++ b/pymc4/distributions/mixture.py
@@ -4,10 +4,8 @@ distributions.
 Wraps tfd.Mixture as pm.Mixture
 """
 
-import collections.abc
-
-import tensorflow as tf
 from tensorflow_probability import distributions as tfd
+
 from pymc4.distributions.distribution import Distribution
 
 
--- a/pymc4/distributions/mixture.py
+++ b/pymc4/distributions/mixture.py
@@ -4,8 +4,8 @@ distributions.
 Wraps tfd.Mixture as pm.Mixture
 """
 
+import collections
 from tensorflow_probability import distributions as tfd
-
 from pymc4.distributions.distribution import Distribution
 
 
@@ -24,12 +24,14 @@ class Mixture(Distribution):
 
     Parameters
     ----------
-    p : array of floats
+    p : array of floats|tensor
         p >= 0 and p <= 1
-        the mixture weights, in the form of probabilities, must sum to one.
-    distributions : multidimensional PyMC4 distribution (e.g. `pm.Poisson(...)`)
-        or iterable of one-dimensional PyMC4 distributions the
-        component distributions :math:`f_1, \ldots, f_n`
+        The mixture weights, in the form of probabilities,
+        must sum to one on the last (i.e. right-most) axis.
+    distributions : PyMC4 distribution|sequence of PyMC4 distributions
+        Multi-dimensional PyMC4 distribution (e.g. `pm.Poisson(...)`)
+        or iterable of one-dimensional PyMC4 distributions
+        :math:`f_1, \ldots, f_n`
     """
 
     def __init__(self, name, p, distributions, **kwargs):
@@ -40,7 +42,7 @@ class Mixture(Distribution):
         p, d = conditions["p"], conditions["distributions"]
         # if 'd' is a sequence of pymc distributions, then use the underlying
         # tfp distributions for the mixture
-        if isinstance(d, (list, tuple)):
+        if isinstance(d, collections.abc.Sequence):
             if any(not isinstance(el, Distribution) for el in d):
                 raise TypeError(
                     "every element in 'distribution' needs to be a pymc4.Distribution object"
@@ -51,6 +53,6 @@ class Mixture(Distribution):
             distr, mixture = d._distribution, tfd.MixtureSameFamily
         else:
             raise TypeError(
-                "'distribution' needs to be a pymc4.Distribution object or a list of distributions"
+                "'distribution' needs to be a pymc4.Distribution object or a sequence of distributions"
             )
-        return mixture(tfd.Categorical(probs=p), distr, **kwargs)
+        return mixture(tfd.Categorical(probs=p, **kwargs), distr, **kwargs)
--- a/pymc4/distributions/mixture.py
+++ b/pymc4/distributions/mixture.py
@@ -96,12 +96,16 @@ class Mixture(Distribution):
                 raise TypeError(
                     "every element in 'distribution' needs to be a pymc4.Distribution object"
                 )
-            distr, mixture = [el._distribution for el in d], tfd.Mixture
+            distr = [el._distribution for el in d]
+            return tfd.Mixture(
+                tfd.Categorical(probs=p, **kwargs), distr, **kwargs, use_static_graph=True
+            )
         # else if 'd' is a pymc distribution with batch_size > 1
         elif isinstance(d, Distribution):
-            distr, mixture = d._distribution, tfd.MixtureSameFamily
+            return tfd.MixtureSameFamily(
+                tfd.Categorical(probs=p, **kwargs), d._distribution, **kwargs
+            )
         else:
             raise TypeError(
                 "'distribution' needs to be a pymc4.Distribution object or a sequence of distributions"
             )
-        return mixture(tfd.Categorical(probs=p, **kwargs), distr, **kwargs)
--- a/pymc4/distributions/mixture.py
+++ b/pymc4/distributions/mixture.py
@@ -4,6 +4,8 @@ distributions.
 Wraps tfd.Mixture as pm.Mixture
 """
 
+import collections.abc
+
 import tensorflow as tf
 from tensorflow_probability import distributions as tfd
 from pymc4.distributions.distribution import Distribution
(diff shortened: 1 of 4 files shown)


twiecki:
Thanks for the contribution! I think the problem with the first implementation is that we're not sampling the non-yielded distributions as the `pm.Model` doesn't know about them. Otherwise I like the API.

lucianopaz:
Thanks for contributing @dirmeier! I'll review what you did later today but I wanted to comment on the two distribution implementations that you talked about. The manual one is wrong.

To see it, I find that it helps to try to write down the generative process with math.

A Mixture distribution can be written as

$$
p = some array
Distributions = a list of distributions
label ~ Categorical(p)
x ~ Distributions[label]
$$

In your first example and mixture distribution implementation, the generative process is equal to

$$
p = some array
m0 ~ Normal(0, 1)
m1 ~ Normal(0, 1)
Distributions[0] = Normal(m0, 1)
Distributions[1] = Normal(m1, 1)
label ~ Categorical(p)
obs ~ Distributions[label]
$$

where I've expanded the Mixture distribution into the explicit sampling from the latent label followed by sampling one of the component distributions. Your manual implementation does the following:

$$
p = some array
m0 ~ Normal(0, 1)
m1 ~ Normal(0, 1)
comp0 ~ Normal(m0, 1)
comp1 ~ Normal(m1, 1)
Distributions[0] = Normal(comp0, 1)
Distributions[1] = Normal(comp1, 1)
label ~ Categorical(p)
obs ~ Distributions[label]
$$

As you can see, you've added an extra layer of samples drawn from gaussian distributions in the second implementation of the model. This makes the model become unidentifiable because m0 and comp0 (and also m1 and comp1) aren't well determined separately, but m0+comp0 is.

Going back to the mixture distribution implementation, we shouldn't create the mixture components inside of the `Mixture` initialization, we should always grab the reference from `component._distribution` (like what you did in your first implementation). Furthermore, it would be nice to have some automatic way of using either `tfd.Mixture` or [`tfd.MixtureSameFamily`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/MixtureSameFamily) depending on the input.

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,36 @@ class Mixture(Distribution):
...
 
     def __init__(self, name, p, distributions, **kwargs):
         super().__init__(name, p=p, distributions=distributions, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions, **kwargs):
+        p, d = conditions["p"], conditions["distributions"]
+        # if 'd' is a list of pymc distributions, then use the underlying
+        # tfp distributions for the mixture
+        if isinstance(d, list):
```
lucianopaz, line 43: We could replace test against a sequence instead of just a list. That way tuples and other array like instances can be accepted

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,36 @@ class Mixture(Distribution):
...
+            distributions = [el._distribution for el in d]
```
lucianopaz, line 44: In this branch, we should always test that every element of `d` is a `pm.Distribution`and raise a `TypeError` otherwise.

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,36 @@ class Mixture(Distribution):
...
+        # if 'd' is a pymc distribution with batch_size > 1, then build
+        # K tfp distributions for tfd.mixture
+        else:
+            # get class of tfd distribution, i.e. Normal/Poisson/etc.
+            ty = type(d._distribution)
```
lucianopaz, line 49: I don't agree with this branch of the if statement. We should
1. Test that `d` is a `pm.Distribution`
2. If the test passes, feed it into `MixtureSameFamily`
3. Raise a `TypeError` otherwise

No intermediate tfd.Distribution other than the mixture and the Categorical should be created in this function.

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,36 @@ class Mixture(Distribution):
...
         or iterable of one-dimensional PyMC4 distributions the
         component distributions :math:`f_1, \ldots, f_n`
     """
 
     def __init__(self, name, p, distributions, **kwargs):
         super().__init__(name, p=p, distributions=distributions, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions, **kwargs):
+        p, d = conditions["p"], conditions["distributions"]
```
lucianopaz, line 40: I'm not completely sure if we should to some safety checks on `p` and `d`'s shape, but I guess that we should trust that tfp will deal with that problem for us when we create the tfd.Mixture*.

pymc4/distributions/mixture.py
dirmeier: Oh, that is way more elegant. However, if only `Categorical` and `Mixture` should are created, that means that a list of distributions of the same type, say `[pm.Poisson(), pm.Poisson(), pn.Poisson()` shouldn't be coalesced to a single Poisson which would allow to use `MistureSameFamily`, right?

dirmeier:
Hey @twiecki and @lucianopaz , thanks for the feedback. 

@lucianopaz , I've implemented on the requested changes.  I added a few test cases as well, but I am not sure if they are sufficient. Could you please check those ?

pymc4/distributions/mixture.py
lucianopaz: Yes, if the user feeds in a list of distributions, we treat them as if they weren't of the same family. We only defer to the `MixtureSameFamily` if the user supplies a distribution instance.

pymc4/distributions/mixture.py
```diff
@@ -1,7 +1,14 @@
+"""PyMC4 Distribution of a random variable consisting of a mixture of other
+distributions.
+
+Wraps tfd.Mixture as pm.Mixture
+"""
+
+from tensorflow_probability import distributions as tfd
+
```
lucianopaz, line 8: Minor nitpick, you can remove this blank line

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,31 @@ class Mixture(Distribution):
     ----------
     p : array of floats
         p >= 0 and p <= 1
-        the mixture weights, in the form of probabilities
+        the mixture weights, in the form of probabilities, must sum to one.
```
lucianopaz, line 29: A key thing that we need to document here is that `p` has to sum 1 over a very specific axis. We need to specify which axis has to sum to one (I don't remember exactly but I think that tfp uses the last one). Also, we could change the type of `p` to something more appropriate, like a `tf.Tensor` and add a type hint for mypy maybe.

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,31 @@ class Mixture(Distribution):
...
 
     def __init__(self, name, p, distributions, **kwargs):
         super().__init__(name, p=p, distributions=distributions, **kwargs)
+
+    @staticmethod
+    def _init_distribution(conditions, **kwargs):
+        p, d = conditions["p"], conditions["distributions"]
+        # if 'd' is a sequence of pymc distributions, then use the underlying
+        # tfp distributions for the mixture
+        if isinstance(d, (list, tuple)):
```
lucianopaz, line 43: You should use something like `collections.abc.Sequence` instead

pymc4/distributions/mixture.py
```diff
@@ -19,11 +26,31 @@ class Mixture(Distribution):
...
+                raise TypeError(
+                    "every element in 'distribution' needs to be a pymc4.Distribution object"
+                )
+            distr, mixture = [el._distribution for el in d], tfd.Mixture
+        # else if 'd' is a pymc distribution with batch_size > 1
+        elif isinstance(d, Distribution):
+            distr, mixture = d._distribution, tfd.MixtureSameFamily
+        else:
+            raise TypeError(
+                "'distribution' needs to be a pymc4.Distribution object or a list of distributions"
```
lucianopaz, line 54: Minor nit, `or a list` should be replaced by `or a sequence`

tests/test_distributions.py
```diff
@@ -546,3 +546,39 @@ def test_extra_parameters(tf_seed, distribution_extra_parameters):
     if distribution_name not in ["Flat", "HalfFlat"]:
         # Test that a sample can be drawn using the alternative extra parameters values
         dist.sample()
+
+
+@pytest.fixture(scope="function", ids=str)
```
lucianopaz, line 551: Ok, I see that we have to do many mixture specific tests so it looks like they should be moved to their own test script. Could you also test the prior and posterior predictive sampling, and also check if a multidimensional `p` and a batch of distributions work too (something like the `multidim_parameters` that we test in other distributions).

pymc4/distributions/mixture.py
dirmeier: `collections.abc.Sequence` recognizes a string as a sequence, too. Shall I still change it?

lucianopaz: Yes, a string, even though it is a sequence, would raise a TypeError later because its elements aren't pm.Distribution instances

tests/test_distributions.py
dirmeier: Hey @lucianopaz , I've implemented some more tests for the mixture class, but some issues came up for the prior/posterior checks regarding shapes which I need to dig into.

I'm not sure though if this is a bug on my side or on TFP's side, since I there's similar problems if I implement the mixture in plain TFP (here's a [colab](https://colab.research.google.com/drive/1XdzhIUANslwKaAYtHUbEBvxUaAFyWFH9?usp=sharing) showing the issue), so any tips would be appreciated.

pymc4/distributions/mixture.py
dirmeier: Right, I added type hints for the probabilities as well as for the distribution. Should I unify this for the other distributions as well (I mean types should be tensors instead of `np.array`s or so).

tests/test_distributions.py
lucianopaz: It looks like a potentially painful issue that vmap runs into when it draws a batch of labels and tries to index into the mixture components list. Could you try passing `use_static_graph=True` to see if that solves the problem?

lucianopaz: @junpenglao, could you take a look at the error that is popping out in the colab? Maybe it's worth opening an issue on tfp?

dirmeier: @junpenglao might you have a look a this?

rrkarim: @dirmeier can you also please consider adding tests with `xla=True`. (I'm not sure if we should flag it as `xfail` though)

dirmeier: Sure :)

rrkarim: @dirmeier are there any issues on pymc4 side still? Disabling dynamic sampling of mixture elements as @lucianopaz recommends, seems to resolve the issues with sampling and the second issue is more related to the tfp coroutines (maybe opening a separate issue in tfp?).

dirmeier: Hi, sorry for the delay. If one sest `use_static_graph` sample drom prior/posterior predictive distributions works. But it would need to be hardcoded in `_init_distribution` as you cant set it via `kwargs`.

rrkarim: Maybe we can add an additional logic in prior/posterior predictive to modify mixture distribution instance attribute (or check whenever it is `use_static_graph=True`) after the model is evaluated. Or maybe overriding `proceed_distribution` to modify the instance attribute.

dirmeier: I don't have a preference here. I think using a static graph per default when instantiating the mixture is probably the easiest and cleanest option. @lucianopaz what do you think on this?

dirmeier: ping @lucianopaz

lucianopaz: @dirmeier, sorry for the delay, I kind of missed what was going on here. Did you finally push some version where you used the static graph? Did it solve the problem?

lucianopaz:
It looks like we have to mark xla as an expected failure in the test suite

dirmeier:
CI finally works now. However, it requires hard coding `use_static_graph` when creating the mixture.

twiecki:
This is great, thanks @dirmeier!

PR #290, opened 2020-06-29
#194 Centralize Fixtures

christhorn2:
This is my first ever Open Source or Python PR, so don't worry about criticising it or even ignoring it - I'm expecting the worst. 

(In fact, it's already been a useful exercise for me so I'm only really raising a PR in case anyone thinks its useful enough to keep or anyone fancies schooling a newbie.)

I figured I'd start by picking up one of the "test" issues as it's a good place to get a high level overview of what's in the codebase. 

The diff isn't particularly clear I'm afraid - since the issue is mainly moving stuff around and refactoring - the latter of which I have tried to keep to a minimum until I get a better understanding!

The rough aim as far as I understood it, is to organise the test fixtures, moving them out of the test files to make those easier to digest and combining any fixtures that might be useful / re-useable and have them in some central place.

To that end: the approach I took was:
- Move the fixtures as they are into a corresponding file located in a new subdirectory called "fixtures". I have prefaced each file with "fixtures" too to reduce duplicate file names. 
- This includes the fixtures that were moved to conftest.py from "test_sampling.py" which I moved into "fixtures_sampling.py"
- I would imagine that many of the simple models can be heavily reused, so fixtures like "simple_model" I have moved into their own file called "fixtures_models.py"
- Quite a few fixtures were previously duplicated in both "test_sampling" and "test_forward_sampling", so I tried to combine those and went for a single file called "fixtures_sampling"
- A few fixtures are different but had the same name, so I renamed them
- I also removed the _fixture suffix on some of them so they are consistent and I feel the test code reads better without it.
- Common (at least as far as I can guess) fixtures like "useXla" I have left in conftest.py
- There a few places I need help - which I have added comments for below.

To the best of my knowledge, I have not changed any of the test functionality. That being said, I did actually add one very small test which is commented below, because it is either missing, or more likely I have something to learn about the framework.

The contributing guidelines suggest I should run "make" which didn't seem to work in my vscode container checkout, but "pytest" passes which I hope is sufficient.

As I said, feel free to ignore but any feedback is welcome.


tests/fixtures/fixtures_executor.py
```diff
@@ -0,0 +1,170 @@
...
+    def model():
+        norm = yield dist.Normal("norm", 0, 1)
+        abs_norm = yield dist.Deterministic("abs_norm", tf.abs(norm))
+        sine_norm = yield dist.Deterministic("sine_norm", tf.sin(norm))
+        norm_copy = yield dist.Deterministic("norm_copy", norm)
+        obs = yield dist.Normal("obs", 0, abs_norm)
+
+    return model, expected_deterministics, expected_ops, expected_ops_inputs
+
+
```
christhorn2, line 128: A challenge for me

tests/fixtures/fixtures_models.py
```diff
@@ -0,0 +1,160 @@
...
+    @pm.model()
+    def simple_model_no_free_rvs():
+        norm = yield pm.Normal("norm", 0, 1, observed=1)
+        return norm
+
+    return simple_model_no_free_rvs
+
+
+# Previously called simple_model in tests_variational
+# Suggest this could be called something better?
```
christhorn2, line 63: Request for help

tests/fixtures/fixtures_models.py
```diff
@@ -0,0 +1,160 @@
...
+    mu = np.zeros(2, dtype=np.float32)
+    cov = np.array([[1, 0.8], [0.8, 1]], dtype=np.float32)
+
+    @pm.model
+    def bivariate_gaussian():
+        density = yield pm.MvNormal("density", loc=mu, covariance_matrix=cov)
+        return density
+
+    return bivariate_gaussian
+
```
christhorn2, line 118: Related to challenge above

tests/fixtures/fixtures_sampling.py
```diff
@@ -0,0 +1,175 @@
...
+def forward_sampling_core_shapes():
+    core_shapes = {
+        "model/mu": (4,),
+        "model/scale": (),  # for forward sampling only
+        "model/x": (5, 4), # for forward sampling only
+    }
+    return core_shapes
+
+# Previously called model_fixture in test_forward_sampling.py
+# Suggest could be called something better?
```
christhorn2, line 76: Request for help

tests/fixtures/fixtures_sampling.py
```diff
@@ -0,0 +1,175 @@
...
+        sd = yield pm.HalfNormal("sd", 1.0)
+        mu = yield pm.Deterministic("mu", tf.convert_to_tensor(1.0))
+        x = yield pm.Normal("x", mu, sd, observed=observed)
+        y = yield pm.Normal("y", x, 1e-9)
+        dy = yield pm.Deterministic("dy", 2 * y)
+
+    return model, observed
+
+# Previously called model_with_observed_fixture in test_forward_sampling.py
+# Suggest could be called something better?
```
christhorn2, line 92: Request for help

tests/test_distributions.py
```diff
@@ -1,490 +1,12 @@
...
-def check_broadcast(broadcast_distribution, request):
-    conditions = _check_broadcast[broadcast_distribution]
-    batch_stack = conditions.pop("batch_stack", (1, 2))
-    event_stack = conditions.pop("event_stack", (3, 4))
-    samples = conditions.pop("samples")
-    dist_class = getattr(pm, broadcast_distribution)
-    dist = dist_class(name=broadcast_distribution, batch_stack=batch_stack, event_stack=event_stack)
-    return dist, samples
-
+from .fixtures.fixtures_distributions import distribution, distribution_conditions, distribution_extra_parameters, broadcast_distribution, check_broadcast
```
christhorn2, line 9: Request for help - what is the "right" way of importing lots of these? I tried * but that came with a depreciation warning.

tests/test_executor.py
```diff
@@ -9,198 +9,28 @@
...
     """Test that model can be defined as method in an object definition"""
-    _, state = pm.evaluate_model(class_model.class_model_method())
-    assert "class_model_method/n" in state.untransformed_values
+    _, state = pm.evaluate_model(simple_model_class.class_model_method())
+    assert "class_model_method/norm" in state.untransformed_values
     assert not state.observed_values
     assert not state.transformed_values
 
-
+## This one I added for completeness - it might not make sense
```
christhorn2, line 23: Definitely review this one!

tests/test_gp_cov.py
```diff
@@ -59,9 +16,10 @@ def build_class_and_get_test_points(name, kwargs):
     return test_points, expected_matrix, expected_point, feature_ndims, KernelClass
 
 
-def test_cov_funcs_matrix_shape_psd(tf_seed, get_data, get_unique_cov_func):
+def test_cov_funcs_matrix_shape_psd(tf_seed, get_data, get_all_cov_func, get_unique_cov_func):
```
christhorn2, line 19: get_all_cov_func <- I think I can do better as per my "hacky" comment above

tests/test_distributions.py
rrkarim: Since we have a directory scope for `conftest.py`. Can we just create a directory hierarchy for the test functions, i.e. separate directories for sampling, execution, gp tests. Each directory will have its own `conftest.py` and the python file (`fixtures.py` for example) of the fixtures. Then to import all the fixtures from `fixtures.py` we can just load a plugin (which in our case is just a python file `fixtures.py`). Something like this:

```python
# our tests/sampling/conftest.py
pytest_plugins = [                                                        
     "fixtures", # fixtures are in tests/sampling/fixtures.py
]
```
Directory scoping also will help us to avoid any ambiguity in fixtures we could have in future. Plus depending on the directory we can have separate logic for hooks.

ColCarroll:
Hey! Thanks for this contribution. It is a lot of changes (especially for a first contribution!), but most of it seems to be shuffling code around. The first thing I would do is get the tests passing again: that would demonstrate that the code still works, and the same code is covered as before. 

Right now the failures are related to running `black`: check out the [contributing instructions](https://github.com/pymc-devs/pymc4/blob/master/CONTRIBUTING.md#step-by-step-instructions). In particular, `make black` and `make check` should be run locally. Let me know if it is confusing or out of date!

I guess the question here is: will centralizing fixtures like this help long run? Three choices are:

- status quo: a `tests/` directory with fixtures where they're needed
- this rearrangement: `tests/` directory, with fixtures in a `fixtures/` directory
- @rrkarim's suggestion: have the `tests/` directory mirror the source directory, with `fixtures.py` in each (let me know if I'm wrong about this!)

We could also put tests next to the source code, though that seems like a bigger change. The second two options organize the code better now, but I feel like we will pay some price in keeping fixtures organized this way. @canyon289 might have an opinion on tests and fixtures?

rrkarim:
> Hey! Thanks for this contribution. It is a lot of changes (especially for a first contribution!), but most of it seems to be shuffling code around. The first thing I would do is get the tests passing again: that would demonstrate that the code still works, and the same code is covered as before.
> 
> Right now the failures are related to running `black`: check out the [contributing instructions](https://github.com/pymc-devs/pymc4/blob/master/CONTRIBUTING.md#step-by-step-instructions). In particular, `make black` and `make check` should be run locally. Let me know if it is confusing or out of date!
> 
> I guess the question here is: will centralizing fixtures like this help long run? Three choices are:
> 
>     * status quo: a `tests/` directory with fixtures where they're needed
> 
>     * this rearrangement: `tests/` directory, with fixtures in a `fixtures/` directory
> 
>     * @rrkarim's suggestion: have the `tests/` directory mirror the source directory, with `fixtures.py` in each (let me know if I'm wrong about this!)
> 
> 
> We could also put tests next to the source code, though that seems like a bigger change. The second two options organize the code better now, but I feel like we will pay some price in keeping fixtures organized this way. @canyon289 might have an opinion on tests and fixtures?

Yes, I say having pytest modules would be the first option to make tests somehow structured for now.

christhorn2:
Thank you both for your feedback. 

> In particular, make black and make check should be run locally. Let me know if it is confusing or out of date!

I couldn't get these commands to work for me locally as `make` wasn't available, I'm probably doing something stupid so I will persevere here and come back if I get stuck.

> I feel like we will pay some price in keeping fixtures organized this way. 

This occurred to be whilst doing the shuffling. One thing I noticed was that this issue was partially addressed [here](https://github.com/pymc-devs/pymc4/pull/238) whereby the fixtures from tests_sampling.py where moved to conftest.py. If the status quo is the desired approach here, should the fixtures be moved back to their original home in tests_sampling?

@rrkarim's suggestion - this sounds cleaner organizationally than my proposal. It's likely I've misunderstood (and if so please help me!) but it's not clear to me that this would nicely reduce duplicate code, which I took to be a part of Issue 194. 

For example, a lot of the duplication here occurred in the `vectorized_model_fixture` used by both test_sampling.py and tests_forward_sampling.py, whose corresponding source files live in different directories. 

If reducing dupe code isn't important, I don't think I would recommend option 2 (this PR).

christhorn2:
I came across [this](https://github.com/PyCQA/pylint/issues/3493) whilst attempting to fix the pylint errors, implying that fixtures should either be in the appropriate test file, or in conftest,

tests/fixtures/fixtures_models.py
Sayam753: I think it will a good idea to rename that `simple_model` fixture to `conjugate_normal_model`.

PR #302, opened 2020-07-24
Revert #271. Distributions now use their conditions attribute instead of _distribution

lucianopaz:
Closes #299.

This PR basically reverts #271. The problem was that the tfp meta distributions: `Sample`, `Independent` and `BatchStacker` would get in the way of the raw underlying `Distribution`, and block direct access to its parameters such as `low` and `high` values.

By the way, I was checking the tests locally and came across several broken doctests. I thought that we were testing those in our CI pipes, but it looks like we aren't. @tirthasheshpatel, could you try to run `pytest -v pymc4 --doctest-modules` and take a look at the error that is popping out from `gp.LatentGP.conditional` and `gp.LatentGP.prior`? I fixed everything else but wasn't able to see the what was wrong with those.

PR #303, opened 2020-07-25
Fix docs in GP submodule

tirthasheshpatel:
As pointed out in #302 by @lucianopaz, I have changed the docs of the GP submodule to pass the `pytest` tests.


lucianopaz:
Thanks @tirthasheshpatel

PR #305, opened 2020-07-27
add jitter argument to gp models

tirthasheshpatel:
From a discussion with @Sayam753, cholesky decomposition fails with `tf.float64` datatype when `tf.math.nextafter` is used. This PR removes this from `stabilize` function and instead lets user choose the amount of noise to add via a `jitter` argument in `prior` and `conditional` methods.

### DIFF
--- a/pymc4/gp/gp.py
+++ b/pymc4/gp/gp.py
@@ -1,6 +1,6 @@
 """An Interface for creating Gaussian Process Models in PyMC4."""
 
-from typing import Union
+from typing import Union, Optional
 
 import tensorflow as tf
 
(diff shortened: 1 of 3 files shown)


pymc4/gp/gp.py
```diff
@@ -250,13 +296,13 @@ def conditional(
         >>> gp = pm.gp.LatentGP(cov_fn=cov_fn)
         >>> @pm.model
         ... def gp_model():
-        ...     f = yield gp.prior('f', X)
-        ...     fcond = yield gp.conditional('fcond', Xnew, given={'f': f, 'X': X})
+        ...     f = yield gp.prior('f', X, jitter=1e-3)
+        ...     fcond = yield gp.conditional('fcond', Xnew, given={'f': f, 'X': X}, jitter=1e-3)
```
AlexAndorra, line 300: Is the `jitter` here related to the one mentionned by the docstrings? If yes, I'm guessing this should be 1e-4 or 1e-6 instead of 1e-3, shouldn't it?

tirthasheshpatel: The docs kept failing with `1e-4` jitter so I had to change it to `1e-4`. Though this isn't related to the default value of `jitter` argument. This is just an independent example using a particular value of `jitter`. So, default values are still `1e-4` for `float32` datatype and `1e-6` for `float64` datatypes. Hope this explains it all...

AlexAndorra: Ah I see, thanks!

tirthasheshpatel:
requesting @AlexAndorra to merge (if no other comments or changes required...).

AlexAndorra:
We gave it 24 hours --> merging 🎉

PR #306, opened 2020-07-28
Compound/Gibbs step and discrete suppot

rrkarim:
- [x] Add tests
- [x] Add Gibbs
- [x] Check on the performance for the `one_step` in compound sampler
- [x] Improve user design. First, make `sampler_type` the default for the model with discrete and continuous distributions.
- [x] Add logging as in pymc3
- [x] Fix seed
- [x] Fix the bug with merge of proposal funcs


rrkarim:
Here are the examples from
```python
@pm.model
def categorical():
    var1 = yield pm.Categorical("var1", probs=[0.2, 0.4, 0.4])
    var1 = yield pm.Categorical("var2", probs=[0.1, 0.3, 0.2, 0.2, 0.2])
    var1 = yield pm.Categorical("var3", probs=[0.1, 0.1, 0.1, 0.1, 0.1, 0.5])

```

rrkarim:
Som results with different proposal generations for `Poisson`. I'm note sure about the poisson proposal generation though. Now it will be able to discover proper samples when scale is provided.
```python
@pm.model
def model():
    var1 = yield pm.Poisson("var1", 4) # gaussian with round dist proposal
    var1 = yield pm.Poisson("var2", 4) # poisson dist proposal
    var2 = yield pm.Poisson("var3", 4) # continous space

trace = pm.sample(model(), 
                  sampler_type="compound", 
                  sampler_methods=[
                      ("var1", RandomWalkM, {"new_state_fn": gaussian_round_fn}),
                      ("var2", RandomWalkM, {"new_state_fn": functools.partial(poisson_fn, scale=4)}), # not sure if this is the right way to define proposal
                  ],
                  trace_discrete=["var1", "var2"])


```

image

pymc4/distributions/state_functions.py
```diff
@@ -0,0 +1,161 @@
+import tensorflow as tf
+import tensorflow_probability as tfp
+from tensorflow_probability.python.mcmc.internal import util as mcmc_util
+
+tfd = tfp.distributions
+
+__all__ = ["categorical_uniform_fn", "bernoulli_fn", "gaussian_round_fn", "poisson_fn"]
```
junpenglao, line 7: You should remove the `poisson_fn`, as the proposal density is not symmetric which makes working with it more difficult. As implemented right now it is not valid.

pymc4/distributions/state_functions.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+
+    def _fn(state_parts, seed):
+        with tf.name_scope(name or "categorical_uniform_fn"):
+            scales = scale if mcmc_util.is_list_like(scale) else [scale]
+            if len(scales) == 1:
+                scales *= len(state_parts)
+            if len(state_parts) != len(scales):
+                raise ValueError("`scale` must broadcast with `state_parts`")
+            deltas = [
```
junpenglao, line 33: Using a tfd.Uniform will be more efficient to draw uniform categorical sample.

pymc4/distributions/state_functions.py
rrkarim: Yeah, I've commented that above. Ok, I will just remove this proposal for now. Poisson proposal is implemented in pymc3 but not implemented as I understand.

pymc4/distributions/state_functions.py
rrkarim: I see, ok fixing this too.

rrkarim:
@junpenglao Can you also suggest something better with the `trace_discrete` argument I've added to `sample()`. I don't like any additional argument, but it is the only way I can have discreteness in the plot for Poisson (see the Picture above). (`tfd.Poisson` can sample dtype `tf.float32` points only, so delta should also be the same dtype)

junpenglao:
@rrkarim is this ready for review? if so could you resolve the conflicts?

rrkarim:
> @rrkarim is this ready for review? if so could you resolve the conflicts?

Done, fixing tests.

pymc4/distributions/state_functions.py
```diff
@@ -0,0 +1,250 @@
...
+    def _fn(self, state_parts: List[tf.Tensor], seed: Optional[int]) -> List[tf.Tensor]:
+        with tf.name_scope(self._name or "categorical_uniform_fn"):
+            deltas = tf.unstack(
+                tf.map_fn(
+                    lambda x: tfd.Categorical(logits=tf.ones(self.classes)).sample(
+                        seed=seed, sample_shape=tf.shape(x)
+                    ),
+                    tf.stack(state_parts),
+                )
+            )
```
junpenglao, line 103: IIUC, this is want you should do. In general, try to avoid using `map_fn` as it is not very efficient.
```suggestion
            deltas = tf.nest.map_structure(lambda x: tfd.Categorical(logits=tf.zeros(self.classes, x.dtype)).sample(
                        seed=seed, sample_shape=tf.shape(x)
                    ), state_parts)
```

pymc4/distributions/state_functions.py
```diff
@@ -0,0 +1,250 @@
...
+class BernoulliFn(Proposal):
+    """
+    Bernoulli proposal sub-class with the `_fn` that is sampling new proposal
+    from bernoulli distribution with p=0.5.
+
+    Parameters
+    ----------
+    scale: Union[List[Any], Any]
+        a `Tensor` or Python `list` of `Tensor`s of any shapes and `dtypes`
+        controlling the scale of the proposal distribution.
```
junpenglao, line 119: Why is scale needed in Bernoulli random?

pymc4/distributions/state_functions.py
```diff
@@ -0,0 +1,250 @@
...
+            state_parts_st = tf.cast(state_parts_st, dtype=scales.dtype)
+            deltas = tf.unstack(
+                tf.map_fn(
+                    lambda x: generate_new_values(
+                        x[0], x[1]
+                    ),  # TODO: some issues with unpack and tf graph
+                    tf.stack([state_parts_st, scales], axis=1),
+                )
+            )
+            return tf.unstack(tf.cast(deltas, dtype=orig_dtype))
```
junpenglao, line 178: Could you walk me through why scale is needed here? Because otherwise it is just a Bernouli random with probs=.5 and shape=state_part

pymc4/distributions/state_functions.py
```diff
@@ -0,0 +1,250 @@
...
+            shape_ = (1,) + shape_[1:] + (shape_[0],)
+            scales = tf.tile(tf.broadcast_to(scales, shape_), inds_tile)[..., 0]
+
+            deltas = tf.unstack(
+                tf.map_fn(
+                    lambda x: generate_new_values(x[0], x[1]),
+                    tf.stack([state_parts_st, scales], axis=1),
+                )
+            )
+            return tf.unstack(deltas)
```
junpenglao, line 242: The shape handling looks overly complicated here, how about something like:
```
def rounded_normal(state_part, scale):
    delta = tfd.Normal(0.0, tf.ones_like(state_part) * scale).sample(
                    seed=seed)
                )
    return tf.cast(delta, state_part.dtype) + state_part
new_state = tf.nest.map_structure(rounded_normal, state_parts, scales)
```

rrkarim: yeap, It could be done easily with `tf.nest.map_structure`, fixing.

rrkarim: Casting won't work because of the line here:
https://github.com/tensorflow/probability/blob/acb010fcc695bdfb74349777b94fa9e5f652375f/tensorflow_probability/python/distributions/poisson.py#L281.
It will crash for `x.dtype = tf.int32`

pymc4/distributions/state_functions.py
rrkarim: Yeap, it is not needed. Sorry, in previous design I had `scale` as the requirement for all proposal functions.

pymc4/distributions/discrete.py
```diff
@@ -296,9 +305,12 @@ class Categorical(BoundedDiscreteDistribution):
     probs : array of floats
         probs > 0 and the elements of probs must sum to 1.
     """
+    _grad_support = False
 
     def __init__(self, name, probs, **kwargs):
         super().__init__(name, probs=probs, **kwargs)
+        classes = probs.shape[-1] if tf.is_tensor(probs) else np.array(probs).shape[-1]
```
junpenglao, line 312: use prefer_static.shape here (it will take care of the if/else automatically).

pymc4/distributions/discrete.py
```diff
@@ -310,7 +322,7 @@ def lower_limit(self):
         return 0.0
 
     def upper_limit(self):
-        return self.conditions["probs"].shape[-1]
+        return float(tf.shape(self.conditions["probs"])[-1])
```
junpenglao, line 325: use prefer_static.shape and also cast to numpy float directly.

pymc4/distributions/state_functions.py
brianwa84: @junpenglao did you mean `tfd.Categorical`?

pymc4/distributions/distribution.py
```diff
@@ -339,12 +344,6 @@ def upper_limit(self):
 class PositiveDiscreteDistribution(BoundedDiscreteDistribution):
     _test_value = 1
 
-    def _init_transform(self, transform):
```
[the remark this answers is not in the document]
rrkarim: fixed

pymc4/inference/sampling.py
```diff
@@ -34,247 +75,141 @@ def sample(
...
-        model,
+    if sampler_type is None:
+        sampler_type = _auto_assign_sampler(model)
+    elif sampler_type == "compound":
+        _log.info("Working with Compound step")
+
+    try:
+        sampler = reg_samplers[sampler_type]
+    except KeyError:
+        print("The given sampler doesn't exist")
```
junpenglao: Use _log.warning instead of print

junpenglao:
This is mostly ready - I suggest we merge this after https://github.com/pymc-devs/pymc4/pull/311.

pymc4/inference/sampling.py
rrkarim: done

junpenglao:
This is awesome contribution! Thanks and great job :-)

PR #285, opened 2020-06-18
ENH: add all covariance functions for gp from PyMC3

tirthasheshpatel:
Covariance functions added:
- [x] `RatQuad`
- [x] `Exponential`
- [x] `Matern52`
- [x] `Matern32`
- [x] `Matern12`
- [x] `Linear`
- [x] `Polynomial`
- [x] `Cosine`
- [x] `Periodic`
- [x] `Gibbs`
- [x] `WarpedInput`
- [x] ~`Coregion`~
- [x] `ScaledCov`
- [x] ~`Kron`~

Tests added:
- [x] `RatQuad`
- [x] `Exponential`
- [x] `Matern52`
- [x] `Matern32`
- [x] `Matern12`
- [x] `Linear`
- [x] `Polynomial`
- [x] `Cosine`
- [x] `Periodic`
- [x] `Gibbs`
- [x] `WarpedInput`
- [x] ~`Coregion`~
- [x] `ScaledCov`
- [x] ~`Kron`~

Docs added:
- [x] `RatQuad`
- [x] `Exponential`
- [x] `Matern52`
- [x] `Matern32`
- [x] `Matern12`
- [x] `Linear`
- [x] `Polynomial`
- [x] `Cosine`
- [x] `Periodic`
- [x] `Gibbs`
- [x] `WarpedInput`
- [x] ~`Coregion`~
- [x] `ScaledCov`
- [x] ~`Kron`~

Other Tasks:
- [x] Add Notebook for mean and covariance functions.
- [x] Fix docs such that the `pydocstyle` lint pass.

Inconsistencies to be addressed as pointed out by @bwengals and @AlexAndorra :
- [x] Finer grid for x values
- [x] Change some parameters to make heat plots look better.

PR #304, opened 2020-07-27
fix docs of kernel functions

tirthasheshpatel:
This PR fixes typos and inconsistencies in the documentation suite of the kernel functions introduced in #285

PR #308, opened 2020-07-29
Fix the doccer tool of kernel functions

tirthasheshpatel:
#285 uses regular expressions to build the docstrings of kernel functions making it slow and prone to errors. I have changed that to build docs using python's standard string interpolation operation.


tirthasheshpatel:
Sorry. closing this as it overwrites the changes of #304

PR #311, opened 2020-08-11
Added pm.Deterministic class objects to sample state

semohr:
Hi,
I tried to overload the pm.*Distributions* classes to add some extra functionality i.e. add some variables which can be defined for plotting or other purposes later.

For example:
```python
class my_Normal(pm.Normal):
    def __init__(self,*args,**kwargs):
        if "foo" in kwargs:
            self.foo = kwargs.get("foo")
        super().__init__(*args,**kwargs)
#And later access it via
_, state = pm.evaluate_model(model_with_my_Normal)

#Now we can access the foo attribute via
state.distributions["model_name/my_Normal_name"].foo #This Works!
```
This is all nice and works but if we try to apply the same concept to the `pm.Deterministic` class we find that `state.deterministics["model_name/my_Normal_name"]` is a tensor and the object is not added to `state.distributions` as well. That is intended behaviour,  so I added an extra dict in the sampling state for deterministic "distributions", which allows me to do:

```python
class my_Deterministic(pm.Deterministic):
    def __init__(self,*args,**kwargs):
        if "foo" in kwargs:
            self.foo = kwargs.get("foo")
        super().__init__(*args,**kwargs)

_, state = pm.evaluate_model(model_with_my_Deterministic)
state.deterministic_distributions["model_name/my_Deterministic_name"].foo #This works now too!
``` 

This all feels a bit hacky and I'm not too sure if this breaks anything but I found it quite useful. If you think this could be a feature feel free to merge the pull request or change anything you see fit. Otherwise just reject/close the request ;)

Best wishes and keep up the good work,
Sebastian

### DIFF
--- a/pymc4/flow/executor.py
+++ b/pymc4/flow/executor.py
@@ -630,6 +630,7 @@ class SamplingExecutor:
                 )
             )
         state.deterministics[scoped_name] = return_value = deterministic.get_value()
+        state.distributions[scoped_name] = deterministic
         return return_value, state
 
     def prepare_model_control_flow(
--- a/pymc4/flow/executor.py
+++ b/pymc4/flow/executor.py
@@ -104,7 +104,7 @@ class SamplingState:
         "distributions",
         "potentials",
         "deterministics",
-        "deterministic_distributions",
+        "deterministic_variables",
     )
 
     def __init__(
@@ -116,7 +116,7 @@ class SamplingState:
         potentials: List[distribution.Potential] = None,
         deterministics: Dict[str, Any] = None,
         posterior_predictives: Optional[Set[str]] = None,
-        deterministic_distributions: Dict[str, distribution.Deterministic] = None,
+        deterministic_variables: Dict[str, distribution.Deterministic] = None,
     ) -> None:
         # verbose __init__
         if transformed_values is None:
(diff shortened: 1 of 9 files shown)


pymc4/flow/executor.py
```diff
@@ -104,6 +104,7 @@ class SamplingState:
         "distributions",
         "potentials",
         "deterministics",
+        "deterministic_distributions",
```
rrkarim, line 107: Could there be a better naming for this attribute though?

fonnesbeck: What is a deterministic distribution?

semohr: I totally agree the name is not well chosen. How about `deterministic_variables`, `deterministic_objects` or  `deterministic_instances` ?

fonnesbeck: More specifically, what's the difference between `determinstics` and `deterministic_distributions`?

semohr: `deterministics` is the tensor i.e. `pm.Deterministic.get_value()` and `deterministic_distributions` is the class instance `pm.Deterministic`

See [L644-L645 ](https://github.com/pymc-devs/pymc4/blob/30b39e411b6a44c6abc9b271c3ee43b7536be342/pymc4/flow/executor.py#L644-L645)

fonnesbeck: In that case, I would change `deterministics` to `deterministics_values`  and `deterministic_distributions` to `deterministics`. That would seem more consistent.

semohr: > In that case, I would change `deterministics` to `deterministics_values` and `deterministic_distributions` to `deterministics`. That would seem more consistent.

Done, I replaced more or less every occurrence of `deterministics` with `deterministics_values`. Was kinda all over the place but should work as far as I can see.

junpenglao:
This looks good to me. @fonnesbeck could we go ahead and merge?

fonnesbeck:
Thanks for this!

PR #313, opened 2020-08-20
Implementation of zero inflated distributions

aj8907:
Hey

I have had a shot at implementing the discrete zero-inflated distributions that were marked as todo, relying on the `tfp` mixture distribution class.

A question I did have was whether  `_test_value` variable needs to be implemented for these classes? 

Feel free to make additional commits to the branch as necessary. Any questions or anything just let me know. 

Thanks


twiecki:
Thanks @aj8907, this looks great. Once tests pass I think we are in good shape.

PR #289, opened 2020-06-29
Add Full Rank Approximation

Sayam753:
This PR adds Full Rank ADVI interface. It is ready to review. Just to experiment with the API, I have created a [gist](https://gist.github.com/Sayam753/23592188b951bdeb53029eb0c4f4f2c3) inspired from @ColCarroll's [notebook](https://nbviewer.jupyter.org/gist/ColCarroll/d673a3af7169bd713bcbdb9445d4a543). I look forward to include mixture comparisons as well.

- [X] Add Full Rank ADVI
- [X] Add docstrings
- [X] Add tests
- [X] Change variable flattening to model flattening
- [X] Add Interval transformation
- [X] Rerun quick start notebook
- [X] Add deterministic callbacks
- [X] Add tests for deterministic callbacks

PR #314, opened 2020-08-23
Remove transformations for Discrete distributions

Sayam753:
As per discussions in #306 and PyMC3's [Discrete distribution](https://github.com/pymc-devs/pymc3/blob/bf8f0bcd86bf1985efa2619370786f6d905ed4da/pymc3/distributions/distribution.py#L217-L219), this PR reverts the addition of transformations to Discrete distributions from #289.

cc @junpenglao


junpenglao:
Thanks!

PR #320, opened 2020-09-08
Add log_likelihood to mcmc and vi samples

### LINKED ISSUE #292: Add log likelihood to arviz trace objects
I tried to run the `loo` function from arviz and got an exception because that part of the model was not annotated in the trace:

```
data_loo = az.loo(tr, pointwise=True)

TypeError: log likelihood not found in inference data object

```

Sayam753:
Resolves #292 

Please review this PR, then I can write tests.

### DIFF
--- a/pymc4/mcmc/samplers.py
+++ b/pymc4/mcmc/samplers.py
@@ -88,6 +88,7 @@ class _BaseSampler(metaclass=abc.ABCMeta):
         seed: Optional[int] = None,
         is_compound: bool = False,
         trace_discrete: Optional[List[str]] = None,
+        include_log_likelihood=False,
     ):
         if state is not None and observed is not None:
             raise ValueError("Can't use both `state` and `observed` arguments")
@@ -158,10 +159,14 @@ class _BaseSampler(metaclass=abc.ABCMeta):
         if len(deterministic_names) > 0:
             posterior.update(dict(zip(deterministic_names, deterministic_values)))
 
+        log_likelihood_dict = dict()
+        if include_log_likelihood:
+            log_likelihood_dict = calculate_log_likelihood(self.model, posterior, state_)
         return trace_to_arviz(
             posterior,
             sampler_stats if is_compound is False else None,
             observed_data=state_.observed_values,
+            log_likelihood=log_likelihood_dict if include_log_likelihood else None,
         )
 
     @tf.function(autograph=False)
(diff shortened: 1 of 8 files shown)


pymc4/mcmc/utils.py
```diff
@@ -127,12 +129,17 @@ def trace_to_arviz(
...
+        }
 
     return az.from_dict(
         posterior=trace,
         sample_stats=sample_stats,
         prior_predictive=prior_predictive,
         posterior_predictive=posterior_predictive,
+        log_likelihood=log_likelihood,
         observed_data=observed_data,
     )
```
Sayam753, line 144: I am a bit confused here whether log_likelihood has to be a group or should it be a variable in `sample_stats`.

Sayam753: As per discussions in ArviZ gitter channel, it's better to have `log_likelihood` as a group for compatibility with Multi-observed random variables. So, I will not change `trace_to_arviz` function.

junpenglao (approved):

Sayam753:
Thanks @junpenglao . I am writing a few tests for this and hopefully push here soon.

pymc4/mcmc/samplers.py
```diff
@@ -822,3 +827,28 @@ def vectorized_logpfn(*state):
...
+
+    def extract_log_likelihood(values, observed_rv):
+        st = flow.SamplingState.from_values(values, observed_values=sampling_state.observed_values)
+        _, st = flow.evaluate_model_transformed(model, state=st)
+        try:
+            dist = st.continuous_distributions[observed_rv]
+        except KeyError:
+            dist = st.discrete_distributions[observed_rv]
+        return dist.log_prob(dist.model_info["observed"])
+
```
Sayam753, line 845: I am not sure here about having deterministic distributions as observed variables. If that's the case, we do need to handle that as well.

Sayam753 (commented):
This PR is all good to go from my side.

Sayam753:
This is interesting. Latest stable release of `nbconvert>=6.0.0` causes the notebooks tests to fail.

mattiasthalen:
When running pm.sample_posterior_predictive, that idata group is not generated. Is that covered by this fix?

Sayam753:
Hi @mattiasthalen 
Sorry for getting back late on this. I checked with both mcmc and vi, that after passing trace (which includes log likelihood) to posterior predictive, the log_likelihood group is still there. 
Can you share a reproducible code snippet, if the issue still persists.
Thanks

mattiasthalen:
> Hi @mattiasthalen
> Sorry for getting back late on this. I checked with both mcmc and vi, that after passing trace (which includes log likelihood) to posterior predictive, the log_likelihood group is still there.
> Can you share a reproducible code snippet, if the issue still persists.
> Thanks

No worries! The mistake was all mine, I forgot to preface the likelihood function with yield... doh! ☺️

PR #321, opened 2020-09-21
Add event_ndims to Transform class to fix shape errors in non-autobatching mode

### LINKED ISSUE #297: Error in transformed variables with manual model vectorization
There seems to be an error when using manual model vectorization for transformed variables when the event shape of the variable is larger than 1. A minimal example would be: 

```python
import numpy as np
import pymc4 as pm
import tensorflow as tf

means = np.random.random((3,1))*5+5
noise = np.random.random((3, 10))

data = means + noise
data = data.astype('float32')


# We want to infer the means of the data:
@pm.model
def model():
    means = yield pm.HalfNormal(name='means', loc= 0, scale = 10, event_stack=3)
    means = tf.repeat(tf.expand_dims(means, axis=-1), axis=-1, repeats=10)
    likelihood = yield pm.Normal(name='likeli', loc=means, scale = 5, observed=data, 
                                 reinterpreted_batch_ndims=2)

trace = pm.sample(model(), num_samples=50, burn_in=200, use_auto_batching=False, num_chains=2)

print(means)
print(np.median(trace.posterior['model/means'], axis=(0,1)))
```

When the `pm.Halfnormal` is replaced by `pm.Normal` it works without problems. 

If I understood the organization of the source code correctly, this error is due to the fact that the correct number of dimensions of the event shape is not passed to `inverse_log_det_jacobian` and `forward_log_det_jacobian` of tensorflow probability, for example exactly here: https://github.com/pymc-devs/pymc4/blob/master/pymc4/distributions/transforms.py#L131. Somehow the Transform class should have also have the number of dimensions of the event_shape as an attribute, to be able to calculate the determinant of the Jacobian correctly.

But eventually, I am using PyMC4 wrongly and there is another way to specify the model...

jdehning:
Closes #297 
### Description of the problem:
When not using auto_batching, the event_shape of the distributions and potentials must match. The reason is, when the log-likelihood is calculated, it is not possible to reduce the sum over all dimensions, as the `batch_shape` has to be preserved. This works well for distributions, however, transformations add a potential as `log_det_jacobian` with too high dimensionality. It is assumed in `SamplingState.collect_unreduced_log_prob` that each potential and distribution consist only of `batch_dim` as shape, however the `backward_log_det_jacobian` also includes the event_dim.

### Solution
An `event_ndims` attribute is added to the transform class that saves the number of event dimensions of the distributions it belongs to. This number is passed to the `log_det_jacobian` function implemented in tensorflow_probability, which then reduces over these axes. At the end of the initialization of a pymc4 distribution, the `event_ndims` attribute of the transform is set by reading out the `event_shape` attribute of the just initialized `tfp.distributions.distribution`. 
There is some additional code to take care of the case when the transformation changes the number of dimensions. This change of dimensions is defined in the `inverse_event_ndims` method of the `tfp.bijectors.bijector`, which is used to calculate the number of dimensions of the transformed variable.

### Comparison to tensorflow_probability
In tensorflow_probability, in `TransformedTransitionKernel`, a similar problem has been already been solved. Their solution was to calculate the rank of the untransformed likelihood function, which is equal to the length of the `batch_shape` and subtract it from the rank of the input to each bijector to get `event_ndims` [link](https://github.com/tensorflow/probability/blob/00a91c4a775e305938fe61e31f1065dda580dfaf/tensorflow_probability/python/mcmc/transformed_kernel.py#L108). This is however done at the sampling stage. I suppose that theoretically, one could also implement something similar in PyMC4 in the SamplingExecutor, however, I think that it is cleaner to do it already at the initialization stage because the logic for the event_shapes of the distributions is also implemented there. 

### Implementation details
* The `event_ndims` is read out from the `tfp.distribution` with prefer_static, which is copied from [link](https://github.com/pymc-devs/pymc4/blob/42642efd89e71787e4d1af6ab8b6ca241b861f35/pymc4/distributions/batchstack.py#L185). 
* One can provide a event_ndims as initialization parameter to a transform object, which than hinders the automatic setting of event_ndims by the distribution class.

### Other Changes
* Fix requirements of xarray <= 0.16.0 because of https://github.com/arviz-devs/arviz/issues/1387
* Fixes some doctests such that the tests run through without errors

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help venv conda docker docstyle format style types black test lint check notebooks
+ .PHONY: help venv conda docker docstyle format style types black test lint check notebooks
 .DEFAULT_GOAL = help
 
 PYTHON = python
@@ -62,8 +62,9 @@ types:
 black:  # Format code in-place using black.
 	black pymc4/ tests/
 
+
 notebooks: notebooks/*
-	jupyter nbconvert --config nbconfig.py --execute --ExecutePreprocessor.kernel_name="pymc4-dev" --ExecutePreprocessor.timeout=1200
+	jupyter nbconvert --config nbconfig.py --execute --ExecutePreprocessor.kernel_name="pymc4-dev" --ExecutePreprocessor.timeout=1200 --to 'html'
 	rm notebooks/*.html
 
 test:  # Test code using pytest.
--- a/pymc4/distributions/distribution.py
+++ b/pymc4/distributions/distribution.py
@@ -5,6 +5,8 @@ from typing import Optional, Union, Any, Tuple
 
 import tensorflow as tf
 from tensorflow_probability import distributions as tfd
+from tensorflow_probability.python.internal import prefer_static
+
 from pymc4.coroutine_model import Model, unpack
 from pymc4.distributions.batchstack import BatchStacker
 from pymc4.distributions import transforms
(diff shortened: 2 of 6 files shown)


jdehning:
There is still an error with the notebooks tests. I don't know exactly how to best fix it

jdehning:
The notebook creation is fixed in the makefile.

